PR #2, opened 2021-09-09
Renaming helm config files for sc-digital-centre

nibivi77:
Just renaming some placeholder text from the template.


shewood (approved):
Looks good

PR #7, opened 2021-09-15
Adding teamcity steps

nibivi77:
## Description

This PR deploys our app to our Dev environment.

[Our dev environment](https://sc-digital-centre.bdm-dev.dts-stn.com/)

Whenever we want to make a change, simply edit the `.teamcity/setting.kts` and make a PR.


ben851 (changes requested):
This is great work! A couple minor comments:
The context environment variable (set to AKS_DEV here) should not indicate a specific environment, since we will use the same files for production as well. Maybe just change it to AKS or back to K8S_CLUSTER_NAME or whatever you want that doesn't indicate a specific environment. 

Second comment, and more just for the back pocket, is that we will want to get that secret init-container back up and running as soon as we start having secrets. That's a whole other topic for discussion though, so keep it like this for now since we'll have to do some solutioning on how we store secrets here.

nibivi77:
> This is great work! A couple minor comments:
> The context environment variable (set to AKS_DEV here) should not indicate a specific environment, since we will use the same files for production as well. Maybe just change it to AKS or back to K8S_CLUSTER_NAME or whatever you want that doesn't indicate a specific environment.
> 
> Second comment, and more just for the back pocket, is that we will want to get that secret init-container back up and running as soon as we start having secrets. That's a whole other topic for discussion though, so keep it like this for now since we'll have to do some solutioning on how we store secrets here.

For the second comment, we were thinking of using KeyVault or Vault for secrets. Once we get there. I am not sure how to do that though.

ben851 (approved):

PR #8, opened 2021-09-17
Adding kts settings to main.

nibivi77:
When changing the settings of the VCS root from TeamCity, our settings.kts file gets overwritten with a new empty one.

this PR adds the settings back in the file? hopefully?


ben851 (approved):
Looks equivalent!

PR #9, opened 2021-09-17
Added metrics and prom-client

DaceyTom2:
### Description

Add logging with Prometheus

Created new endpoint at /api/metrics
Navigate and view metrics

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,10 +5,10 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "next-template",
       "version": "0.1.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


TishamIslam (approved):
Loaded it and saw a bunch of different metrics, so I guess it's working.

nibivi77 (approved):
Tested and functional. Will this be an api we will want to hide?

DaceyTom2:
@nibivi77 We are in the talks of hiding the api endpoint. @ben851 doesn't think it is a big deal for dev but we will have to come up with a solution for prod

PR #10, opened 2021-09-17
Dc 39 prom client

shewood:
### Description
Installed prom-client v 12.0.0 (not latest which is 13)
Added a server.js file. 

List of proposed changes:

-
-


nibivi77 (commented):
I also believe The other PR is a little better than this one. Even if they both do the same thing :p

shewood:
I am closing this PR in favour of Tom's using the api method instead of the server.js file to call the prom client.

PR #11, opened 2021-09-17
Using TC build date as last date modifed.

nibivi77:
There are no tasks for this but I figured it is sort of related to our versioning.

The code is mostly copied from Life-Journeys.

To test it, it needs to be pushed to main. This means that the date will only show on the deployed version (Dev or Prod).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ COPY . .
 
 FROM base AS build
 ENV NODE_ENV=production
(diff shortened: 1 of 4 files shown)


DaceyTom2:
I am getting DATE-NA as the date modified 
image

nibivi77:
> I am getting DATE-NA as the date modified
> image

Yep thats normal. It won't work locally.

MarcoGoC (approved):
this looks good to me 

is TC going to work? will see

PR #12, opened 2021-09-20
DC-36 get data from aem

MarcoGoC:
### Description
"Sample" code to get data from the AEM API (if available) or from a local file 

List of proposed changes:
- Sample code on index,js that gets the data from for the page
- API code file.  

- N/A for now

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+NEXT_PUBLIC_BUILD_DATE = 20210917
+
+NEXT_CONTENT_API = " api address here "
--- a/next.config.js
+++ b/next.config.js
@@ -8,7 +8,7 @@ const builddate = process.env.NEXT_PUBLIC_BUILD_DATE
   : 'DATE-NA'
 
 // AEM base end point
-var contentURL = process.env.NEXT_CONTENT_API
+const contentURL = process.env.NEXT_CONTENT_API
   ? process.env.NEXT_CONTENT_API
   : ''
 
--- a/next.config.js
+++ b/next.config.js
@@ -15,6 +15,7 @@ const contentURL = process.env.NEXT_CONTENT_API
 module.exports = {
   env: {
     NEXT_PUBLIC_BUILD_DATE: builddate,
+    NEXT_CONTENT_API: contentURL,
   },
   reactStrictMode: true,
 }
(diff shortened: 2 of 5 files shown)


next.config.js
```diff
@@ -7,6 +7,11 @@ const builddate = process.env.NEXT_PUBLIC_BUILD_DATE
     process.env.NEXT_PUBLIC_BUILD_DATE.substr(6, 2)
   : 'DATE-NA'
 
+// AEM base end point
+var contentURL = process.env.NEXT_CONTENT_API
```
nibivi77, line 11: If possible, can we change `var` to a `const`?

nibivi77 (approved):

nibivi77:
Looks good to me Marco!

.env.example
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_BUILD_DATE = 20210917
```
DaceyTom2, line 1: .env needs to be added to git ignore

pages/api/getData.js
```diff
@@ -0,0 +1,29 @@
...
+  return { localData: [] }
+}
+
+//
+// Fetch ALL topics from API when available
+//
+
+export async function getTopics(language) {
+  const res = await fetch(
+    `${process.env.NEXT_CONTENT_API}${language}/lj-new-child.json?date=${process.env.NEXT_PUBLIC_BUILD_DATE}`
```
DaceyTom2, line 23: If lj-new-child was a parameter would we be able to make something reusable? Perhaps there is a framework for AEM requests?

DaceyTom2 (changes requested):

.env.example
MarcoGoC: next uses .env.local

pages/api/getData.js
MarcoGoC: we could but it depends on how we implement the AEM structure

.env.example
DaceyTom2: My apologies, I didn't realise this was an example

pages/api/getData.js
DaceyTom2: Sounds good

DaceyTom2 (approved):
Looks good!

PR #13, opened 2021-09-20
Adding build_date arg.

nibivi77:
I think the build time for the last date modified did not work because I forgot to define it in our Dockerfile as an ARG.


MarcoGoC (approved):
Thanks

PR #14, opened 2021-09-20
Adding changelog

nibivi77:
Adding simple changelog file.


TishamIslam (approved):
Looks good, should we be adding to it with every PR and separating new additions by sprints?

nibivi77:
> Looks good, should we be adding to it with every PR and separating new additions by sprints?

Adding a line in the changelog is not necessary. However, we will be separating it by sprints.

PR #15, opened 2021-09-20
added colours from sc labs dc design system

TishamIslam:
### Description
Added the colours listed [here](https://www.figma.com/file/9LVfXvyx75Hh7H6eXrDBL4/Service-Canada-Digital-Centre-Design-System?node-id=5056%3A1429)

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -8,19 +8,16 @@ module.exports = {
         gray: {
           secondary: '#7f8c8d', //paragraphs and prefooter bg
           light: '#cfcfcf',
+          dark: '#262626',
         },
         white: {
           DEFAULT: '#FFFFFF',
           light: '#fbfbfb',
         },
-        black: {
-          DEFAULT: '#000000',
-          dark: '#262626',
-        },
         blue: {
           //primary
           'primary-deep': '#173451',
-          'bright': '##3498db',
+          'bright': '#3498db',
 
           //secondary
           'secondary-deep': '#3a6b90',
@@ -31,7 +28,7 @@ module.exports = {
           bright: '#87d37c',
         },
         red: {
-          alert: '##d94141',
+          alert: '#d94141',
         },
         orange: {
           alert: '#EE7100',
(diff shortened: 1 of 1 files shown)


tailwind.config.js
```diff
@@ -3,7 +3,43 @@ module.exports = {
...
   theme: {
-    extend: {},
+    extend: {
+      colors: {
+        gray: {
+          secondary: '#7f8c8d', //paragraphs and prefooter bg
+          light: '#cfcfcf',
+        },
+        white: {
+          DEFAULT: '#FFFFFF',
```
nibivi77, line 13: I don't think we need to add the white and black DEFAULT as they are already defined in Tailwind.

nibivi77 (approved):
Good!  Check out the comment :)

tailwind.config.js
```diff
@@ -3,7 +3,43 @@ module.exports = {
...
+          light: '#fbfbfb',
+        },
+        black: {
+          DEFAULT: '#000000',
+          dark: '#262626',
+        },
+        blue: {
+          //primary
+          'primary-deep': '#173451',
+          'bright': '##3498db',
```
nibivi77, line 23: I think there's a typo here with 2 #

tailwind.config.js
```diff
@@ -3,7 +3,43 @@ module.exports = {
...
+        gray: {
+          secondary: '#7f8c8d', //paragraphs and prefooter bg
+          light: '#cfcfcf',
+        },
+        white: {
+          DEFAULT: '#FFFFFF',
+          light: '#fbfbfb',
+        },
+        black: {
+          DEFAULT: '#000000',
```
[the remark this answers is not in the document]
TishamIslam: Changed to gray-dark 👍

tailwind.config.js
TishamIslam: I think because tailwind only has "white" and "black", adding them as objects in the config erases the defaults so we have to add them back. Had an error come up while trying to use the colours without the defaults.

PR #16, opened 2021-09-22
Prometheus troubleshooting

ben851:
### Description

List of proposed changes:

- Added service annotations with correct URL and port for Prometheus to automatically scrape the app
- Removed the "root" ingress path in the dev environments (this is reserved for prod only)

helmfile --environment dev apply
verify in bdm-dev grafana that you can see the instance in the "NodeJS Test Dashboard" instance drop down


shewood (commented):
It looks ok to me but I don't know much about Helm files.

shewood (approved):
as per my first comment. It looks ok to me but I have limited knowledge.

DaceyTom2 (approved):
Looks good

PR #17, opened 2021-09-23
Dc 23 splash page

TishamIslam:
### Description
Replaced default index page with splash page. The page looks and behaves more like the VA's splash page at the moment, we can change it later.

List of proposed changes:
- splash page addition
- changed unit test so that it passes on the new splash page. very basic

Does it look good?


nibivi77 (approved):
This is great Tisham! The only thing I would change is the name of the test file from "home" to "index" or "splash"

PR #18, opened 2021-09-23
Footer

nibivi77:
### Description

Added the Footer from Alpha and added some basic tailwind Styles from Alpha as well.
Also added the basic `Layout` component.

I also added the Docker ARG we need for the build time on the deployed version.

Open [Alpha](https://alphasite.dts-stn.com/home) and DC and compare both.


nibivi77:
I might add tests to this PR. It would just be a "make sure it renders" kinda test.

tailwind.config.js
```diff
@@ -3,6 +3,35 @@ module.exports = {
   purge: ['./pages/**/*.{js,ts,jsx,tsx}', './components/**/*.{js,ts,jsx,tsx}'],
   darkMode: false,
   theme: {
+    fontFamily: {
+      display: ['Lato'],
+      body: ['Noto sans'],
+    },
+    fontSize: {
```
shewood, line 10: can we not add these from alpha, it is really messy. Just add what we need when we need it?

tailwind.config.js
```diff
@@ -3,6 +3,35 @@ module.exports = {
...
+      xxs: ['12px', '16px'],
```
shewood, line 11: and this, we didn't need this at all and it's all default in tailwind.

tailwind.config.js
```diff
@@ -3,6 +3,35 @@ module.exports = {
...
+      p: ['20px', '30px'],
+      h4: ['22px', '20px'],
+      h3: ['24px', '24.3px'],
+      h2: ['30px', '33.5px'],
+      h1: ['36px', '42px'],
+      h1l: ['38px', '42px'],
+      h1xl: ['45px', '54px'],
+      h1xxl: ['70px', '84px'],
+    },
+    screens: {
```
shewood, line 25: Also, we should see what screensizes Sidra recommends

tailwind.config.js
```diff
@@ -22,6 +51,10 @@ module.exports = {
           //secondary
           'secondary-deep': '#3a6b90',
           'dull': '#5996b2',
+
```
shewood, line 54: would it be possible to organize the colours in the same way as defined, so gray-light etc as these colours are used elsewhere. Just change in the jsx to match

shewood (changes requested):
Looks good, Just a few comments on importing stuff from alpha. it is really messy and it would be nice to use the tailwind defaults as much as possible and only add new things when we run into an issue where it needs updating to match a specific design.

DaceyTom2 (changes requested):
Looks good, something small

tailwind.config.js
nibivi77: I am using the font family in the footer but I will remove what we don't need.

DaceyTom2 (approved):
Looks good!

PR #19, opened 2021-09-23
DC34 adding rewrites

MarcoGoC:
### Description
Adding rewrites to handle 

List of proposed changes:
- changes to next.config
- home page **place holder** 
- index page links to the 'english' and 'french' version of the page

- download and run 
- click on the 'language' buttons


components/atoms/ActionButton.js
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Button component
+ */
+export function ActionButton(props) {
+  //Styling for buttons and links
+  const basicStyle =
+    'rounded-sm focus:ring-1 focus:ring-black focus:ring-offset-2'
+  const defaultStyle =
+    'py-2 px-4 bg-custom-blue-blue text-white border border-custom-blue-blue active:bg-custom-blue-dark hover:bg-custom-blue-light'
```
TishamIslam, line 13: We don't have all of these colours defined in Tailwind

components/atoms/ActionButton.js
```diff
@@ -0,0 +1,192 @@
...
+        }
+        ${props.secondary && !props.disabled ? secondaryStyle : props.className}
+        ${props.tertiary && !props.disabled ? tertiaryStyle : props.className}
+        ${props.custom && !props.tertiary ? props.custom : ''}
+        ${props.disabled ? disabledStyle : props.className}`}
+        onClick={props.onClick}
+        id={props.id}
+        data-testid={props.dataTestId}
+        data-cy={props.dataCy || props.id}
+        data-cy-button={props.dataCyButton}
```
TishamIslam, line 54: Maybe remove the data fields for now, as they won't be used for a while I think.

TishamIslam (changes requested):
There's a few things in the ActionButton component that we're not using/don't have in the project e.g. tailwind classes for styles, the data fields in the component, etc. I think we could either have them empty, or remove them entirely for now.

components/atoms/ActionButton.js
nibivi77: I would personally leave them because we should use them. even if its in the future.

But thats just me

components/atoms/ActionButton.js
nibivi77: i agree we can remove the unsued colors and styles

shewood: until the colours are resolved, just pick a close colour that is represented by tailwind like blue-200 or something like that. we can refactor them later.

components/atoms/ActionButton.js
shewood: I vote to leave them as there is a lot of functionality in this button that will become useful. maybe some comments might be nice thought to explain what different parts do or are for etc.

TishamIslam: That makes sense, we could also add that they aren't doing anything atm so we ignore it for now if we add more to it

MarcoGoC: ActionButton is not used yet on any page, we should wait until we decide on the colours to start deleting code that we might need later

components/atoms/ActionButton.js
MarcoGoC: ActionButton is not used yet on any page, we should wait until we decide on the colours to start deleting code that we might need later

TishamIslam (approved):
Good work!

PR #20, opened 2021-09-23
extracted btn-primary, added btn styles from design lib

TishamIslam:
### Description
Just made stuff more readable and made it so that it'd reflect the [canada.ca design](https://www.canada.ca/home.html) better 

List of proposed changes:
- [Design library button styles](https://www.figma.com/file/TodbPq5LF1G6l1E3Kx2qPP/GC-Design-Library?node-id=19%3A0) added into the project
- Extracted btn-primary component, can add the other btns as needed

### Addtional Notes
Please take a look at the extracted component and point out anything that should be removed.

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -25,24 +25,19 @@ export default function Splash() {
             <Link href="/en">
               <a
                 className="
-                                      focus:ring-1 focus:ring-black focus:ring-offset-2
-                                      px-4
-                                      bg-blue-primary-deep
-                                      text-white
-                                      border border-blue-primary-deep
-                                      active:bg-blue-dull
-                                      hover:bg-blue-secondary-deep
-                                      text-center text-md
-                                      h-11
-                                      xl:h-10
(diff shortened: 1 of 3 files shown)


MarcoGoC:
I suggest using ActionButton as per alpha

shewood (changes requested):
Maybe hold off on the role based colours until we can discuss with the designers as there are several duplicate colours in the list and it will get confusing later on.

pages/index.js
```diff
@@ -55,23 +50,19 @@ export default function Splash() {
...
+                  h-11
+                  xl:h-10
+                  w-28
+                  xl:w-36
+                  py-3
+                  rounded
+                  leading-3
+                  shadow-md
+                "
                 id="french-button"
```
DaceyTom2, line 66: The buttons padding got a bit wonky
image

DaceyTom2 (changes requested):

PR #22, opened 2021-09-24
Added Breadcrumb and tests.

nibivi77:
### Description

Added breadcrumb component. It is not added to any page as it will be added from the home page and not the index (splash) page.

Also added basic tests.

`npm i` needs to be ran.
Then run `npm run test:unit`

### DIFF
--- a/components/atoms/Breadcrumb.js
+++ b/components/atoms/Breadcrumb.js
@@ -1,0 +1,58 @@
+import PropTypes from 'prop-types'
+import Link from 'next/link'
+
+/**
+ *  Breadcrumb component
+ */
+export function Breadcrumb(props) {
+  return (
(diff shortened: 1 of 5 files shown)


shewood (approved):
This looks good but take note that in your other PR 'bright-anchor': '#0535d2', is the same as what you have added here under blue-link and in the code it looks like there is still a reference to canada-footer-blue or something like that. I would say leave it for now and we can refactor all of the colours once we get finals from design.

nibivi77:
> This looks good but take note that in your other PR 'bright-anchor': '#0535d2', is the same as what you have added here under blue-link and in the code it looks like there is still a reference to canada-footer-blue or something like that. I would say leave it for now and we can refactor all of the colours once we get finals from design.

I changed it in the other PR so that it reflects this one.

PR #28, opened 2021-09-28
Simpler ActionButton

MarcoGoC:
### Description

Simpler ActionButton code

- Removed all the colours but the default 
- Removed unnecessary list of explicit named attributes

### DIFF
--- a/components/atoms/ActionButton.js
+++ b/components/atoms/ActionButton.js
@@ -1,117 +1,58 @@
 import PropTypes from 'prop-types'
 import Link from 'next/link'
-import { useEffect } from 'react'
 
 /**
  * Button component
  */
-export function ActionButton(props) {
+export function ActionButton({
+  href,
+  custom,
+  className,
+  onClick,
+  type,
+  role,
+  draggable,
+  lang,
+  id,
+  text,
+  children,
+  ...rest
+}) {
   //Styling for buttons and links
-  const basicStyle =
-    'rounded-sm focus:ring-1 focus:ring-black focus:ring-offset-2'
   const defaultStyle =
-    'py-2 px-4 bg-custom-blue-blue text-white border border-custom-blue-blue active:bg-custom-blue-dark hover:bg-custom-blue-light'
-  const secondaryStyle =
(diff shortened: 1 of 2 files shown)


components/atoms/ActionButton.js
```diff
@@ -1,117 +1,56 @@
...
-        ${props.tertiary && !props.disabled ? tertiaryStyle : props.className}
-        ${props.custom && !props.tertiary ? props.custom : ''}
-        ${props.disabled ? disabledStyle : props.className}`}
-        onClick={props.onClick}
-        id={props.id}
-        data-testid={props.dataTestId}
-        data-cy={props.dataCy || props.id}
-        data-cy-button={props.dataCyButton}
-        disabled={props.disabled}
+        className={`${!custom ? defaultStyle : className}`}
```
DaceyTom2, line 26: Instead of passing a boolean for custom can we do a check to see if className is defined?
Then instead of passing props custom=true and className=example class we can just pass className and it is inferred to overwrite the default class?

components/atoms/ActionButton.js
```diff
@@ -1,117 +1,56 @@
...
-      ${props.disabled ? disabledStyle : props.className}`}
-      onClick={props.onClick}
-      type={props.type}
-      id={props.id}
-      data-testid={props.dataTestId}
-      data-cy={props.dataCy || props.id}
-      data-cy-button={props.dataCyButton}
-      disabled={props.disabled}
-      data-gc-analytics-submit={props.analyticsTracking ? 'submit' : undefined}
+      className={`${!custom ? defaultStyle : className}`}
```
DaceyTom2, line 40: Instead of passing a boolean for custom can we do a check to see if className is defined?
Then instead of passing props custom=true and className=example class we can just pass className and it is inferred to overwrite the default class?

nibivi77:
This is much better! I do agree with what Tom is saying.

nibivi77 (approved):
I much prefer this Action Button now! :)

DaceyTom2 (approved):
Looks great Marco! Thank you for cleaning up the action button

PR #31, opened 2021-10-04
DC-55 updated colours as per the new design

shewood:
### Description
updated the colours and some styles as per new design system in Figma. 

List of proposed changes:
updated the colours to match the figma palette of colours. 
Some might not yet be correct as the design needs to also be updated to use the requested colours so some changes still may need to be made. 

-
-


TishamIslam (approved):
Looks great overall, up to you if you want to merge right away or wait for the colours to get finalized and if necessary, ask for re-review.

nibivi77 (approved):
Looks good :)

PR #32, opened 2021-10-05
DC 26 header

MarcoGoC:
### Description
Added the header component that contains the language selector and placeholder for search bar, nav menu, and breadcrumbs

List of proposed changes:

- Added Header component to Layout 
- Added Meta component to Layout
- Added test for both components

- Download branch and run


components/molecules/Header.js
```diff
@@ -0,0 +1,130 @@
...
+            </a>
+
+            {/* Language selector for small screens */}
+            <Link
+              key={language}
+              href={router.asPath}
+              locale={language === 'en' ? 'fr' : 'en'}
+            >
+              <a
+                className="visible md:invisible md:text-sm ml-6 sm:ml-16 underline font-body font-bold text-canada-footer-font  text-base hover:text-canada-footer-hover-font-blue"
```
DaceyTom2, line 54: It's better to use display hidden instead of invisible.
https://tailwindcss.com/docs/display

This makes the element not take up space on the page

components/molecules/Header.js
```diff
@@ -0,0 +1,130 @@
...
+
+          <div className="flex-col flex">
+            {/* Language selector for mid to larger screens */}
+            <Link
+              key={language}
+              href={router.asPath}
+              locale={language === 'en' ? 'fr' : 'en'}
+            >
+              <a
+                className="md:visible invisible pb-0 lg:pb-2 self-end underline font-body text-canada-footer-font hover:text-canada-footer-hover-font-blue "
```
DaceyTom2, line 71: Same as above for visibility

components/molecules/Header.js
```diff
@@ -0,0 +1,130 @@
...
+          href="#wb-cont"
+          draggable="false"
+        >
+          {t.skipToMainContent}
+        </a>
+      </nav>
+
+      <header>
+        <div className="layout-container flex-col flex md:flex md:flex-row justify-between mt-2">
+          <div className="flex flex-row justify-between items-center lg:mt-7 mt-1.5">
```
DaceyTom2, line 34: The margin at the top can be reduced a fair bit to match the design better

components/molecules/Header.js
MarcoGoC: https://tailwindcss.com/docs/visibility they do recommend

DaceyTom2: They recommend both but invisible takes up space
image

nibivi77 (approved):
This is great work Marco. Do you think the navigation menu should be in the Header component? or below it in the layout.js?

DaceyTom2:
Looks great!

DaceyTom2 (approved):

PR #33, opened 2021-10-05
Husky commit on lint fixed

DaceyTom2:
### Description

List of proposed changes:

Fixed issue where non staged files were being committed by husky when linting

Update 2 files with bad linting.
Stage 1 of the files.
Commit files without pushing
Unstaged file should not be committed


MarcoGoC:
This changes do the same as removing   "&& git add -A" , doesn't it ??

shewood (approved):
Looks good to me.

DaceyTom2:
@MarcoGoC It still commits the files. The big difference is the lint only applies to staged files

PR #34, opened 2021-10-05
Dc 67 top task component

DaceyTom2:
### Description

Top task design for mobile and desktop

List of proposed changes:

Created component for Top task with an example of how to use in home commented out

### Addtional Notes
Since the container it lives in doesn't exist yet I have it floating on /home commented out.
Currently the data comes in as props but will require AEM


components/molecules/TopTasks.js
```diff
@@ -0,0 +1,40 @@
...
+export default function TopTasks(props) {
+  return (
+    <>
+      <h2 className="font-display font-bold text-xl md:text-2xl text-dark-solid">
+        Top Tasks
+      </h2>
+      <p className="mt-3 font-body md:hidden">{props.topTasksDescription}</p>
+      <ul className="list-disc m-5">
+        {props.topTasksList.map((topTask, index) => (
+          <li className="text-dark-solid">
```
MarcoGoC, line 13: missing key=index

DaceyTom2: Ty!

nibivi77 (approved):
Good! The layout of all our components is going to make this fit better later. Same as for the colors.

PR #35, opened 2021-10-05
Clean LGTM report

DaceyTom2:
### Description

Removed unused imports

### DIFF
--- a/pages/api/getData.js
+++ b/pages/api/getData.js
@@ -1,3 +1,6 @@
+// import fs from 'fs'
+// import path from 'path'
+
 //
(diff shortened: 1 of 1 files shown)


pages/api/getData.js
```diff
@@ -1,6 +1,3 @@
-import fs from 'fs'
-import path from 'path'
```
MarcoGoC, line 2: Maybe comment this lines instead or removing the function that loads data from local files ??

DaceyTom2: Added them as comments :)

nibivi77 (approved):
Cool! thank you

PR #38, opened 2021-10-06
Added most requested section and tests.

nibivi77:
### Description

Added the Most requested section from the Figma:

image
image

Try different screen sizes.


nibivi77:
> Looks good, it might need some margin-left and right. it depends on how the holding container is implemented in the page

Yes, and that is exactly why I did not add any for now.

PR #40, opened 2021-10-07
added cardwithimagelayout component

TishamIslam:
### Description
Added a component that simplifies making layouts using a card alongside an image.

### DIFF
--- a/components/molecules/CardWithImageLayout.js
+++ b/components/molecules/CardWithImageLayout.js
@@ -1,0 +1,18 @@
+import PropTypes from 'prop-types'
+
+export default function CardWithImageLayout(props) {
+  return (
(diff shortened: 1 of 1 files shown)


components/molecules/CardWithImageLayout.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types'
+
+export default function CardWithImageLayout(props) {
+  return (
+    <div className={'flex flex-col sm:flex-row w-full ' + props.layoutClasses}>
+      <div className={props.cardClasses + ' flex '}>{props.card}</div>
```
nibivi77, line 6: I don't know if we need these div's. Seems a little redundant to have div's on top of div's.

An option would be to change this:
image
 
 
 To something like this:
 
image

PR #45, opened 2021-10-08
quick fix to ease linking jira issues

TishamIslam:
### Description
Changed the issue/PR templates so that they're reflective of the new Jira board prefix (from DC to DCWC)


TishamIslam:
Wow that typo went unnoticed for a long while, nice catch!

PR #48, opened 2021-10-12
Removed all unnecessary lang imports.

nibivi77:
Removed all necessary imports for i18n.

Fixed a few tests to make everything pass.

## Notes
-  Layout i18n is kept within layout.js for future pages. 
- In the home.js page, I left the list of items (MostRequested Items and Top Tasks items) hard-coded since they will come from the AEM.

### DIFF
--- a/components/molecules/Header.js
+++ b/components/molecules/Header.js
@@ -3,12 +3,8 @@ import Link from 'next/link'
 import { Menu } from '../molecules/Menu'
 import { useRouter } from 'next/router'
 
-import en from '../../locales/en'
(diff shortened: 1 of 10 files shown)


components/molecules/Header.js
```diff
@@ -123,6 +119,11 @@ Header.propTypes = {
    * current locale in the address
    */
   locale: propTypes.string,
+
+  /**
+   * Translated text
+   */
+  t: propTypes.object,
```
DaceyTom2, line 126: I'm wondering if it is worth passing t in or if we should just pass the text in

PR #49, opened 2021-10-13
Added zap workflow

DaceyTom2:
### Description

ZAP scan added to workflow

List of proposed changes:

- ZAP scan added to workflow

Unable to test without PR


TishamIslam (approved):
Seems to be working fine, considering it automatically opened a new issue with all the things that the ZAP run found. We can further configure it later.

PR #64, opened 2021-11-01
Dcwc 642 modal

TishamIslam:
### Description
Added filter modal component

List of proposed changes:

- Added filter modal
- Added jest tests

### Additional Notes
Uncomment the modal from the filters page, we can add it for realsies once we have some functionality on the page.


DaceyTom2 (approved):
This looks really good. Not sure if it is needed yet but we may want to implement a way to x out or cancel in time

PR #87, opened 2021-11-03
DCWC-678 - Get benefit text from AEM

shewood:
### Description
With help from TOM and Marco, the featured benefit title, description and image are now getting pulled from AEM

List of proposed changes:

### Additional Notes
Note that there are 3 warnings for act(...) that still need to be resolved.


DaceyTom2 (approved):
Looks great!

PR #102, opened 2021-11-05
Dcwc 641 filters component desktop

TishamIslam:
### Description
Added the filters form for the desktop view as a sidebar. Also updated the filter component in general.
 
List of proposed changes:

- Add filters form on desktop view
- Add supplementary tests as necessary
- Rearrange desktop view to fit the new form 
- Added focus trap to the modal element component
- Added i18n language support
- Filter form now uses dynamic lists of options (this is in case we're going to be using AEM to get the options, though it also makes the overall code more concise)


components/molecules/ModalElement.js
```diff
@@ -0,0 +1,41 @@
+import PropTypes from 'prop-types'
+
+export default function SearchFiltersModal(props) {
```
nibivi77, line 3: Can we keep the file naming convention to be the same as the component please?

components/molecules/ModalElement.js
```diff
@@ -0,0 +1,41 @@
+import PropTypes from 'prop-types'
```
nibivi77, line 1: For the modal focus, I highly suggest checking out [focus-trap-react](https://www.npmjs.com/package/focus-trap-react)

I used it before in Benefits Finder [here](https://github.com/DTS-STN/Benefits-Finder/blob/main/components/organisms/Drawer.js)

The problem is that when the modal is open, we can still tab through the stuff in the back.

nibivi77 (changes requested):
Bravo Tisham, honestly well done!

As for all the hardcoded English strings, do we want to leave them there or add them to our locale files?

Few comments are below.

components/molecules/ModalElement.js
TishamIslam: Good catch!

nibivi77 (approved):
This looks good and I will approve but there is one thing to point out.

If you go on Mobile, open the filter, and then stretch the screen to desktop view WITHOUT closing the filter, the app becomes unresponsive since the Modal is still "opened".

Is this a bug that we want to fix before merging?

PR #103, opened 2021-11-05
Dcwc added store

MarcoGoC:
### Description
List of proposed changes:

- Added React.Context to app and the context file to manage the store
-

- download and run 
- you should see the data in the store under the BenefitsContext component

### Additional Notes
image


nibivi77 (approved):
Nice we can now use a store! Thank you :)

PR #104, opened 2021-11-09
added image for search adjust heights of cards

shewood:
## [DCWC-XXX](https://jira-dev.bdm-dev.dts-stn.com/browse/DCWC-XXX) (Jira Issue)
(jira doesn't seem to be working)

### Description
Added an image to the search bar area
Updated the Cards so that they grow in the row to the size of the card with the largest amount of text. 
Had to futz with the button to make it as it was before. 
Tried not to break anything.

### DIFF
--- a/components/organisms/ImageBox.js
+++ b/components/organisms/ImageBox.js
@@ -1,17 +1,11 @@
-import propTypes from 'prop-types'
+import PropTypes from 'prop-types'
 
 import Image from 'next/image'
 
-import SearchCard from '../molecules/SearchCard'
-
-import en from '../../locales/en'
-import fr from '../../locales/fr'
 /**
  *
  */
 export default function ImageBox(props) {
-  const t = props.locale === 'en' ? fr : en
-
   return (
     <div className=" bg-deep-blue-solid">
       <div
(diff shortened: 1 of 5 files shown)


components/organisms/ImageBox.js
```diff
@@ -0,0 +1,50 @@
...
+          paraText={t.searchDesc}
+          viewBenefitsServices={t.searchViewAllBenefits}
+          searchBarPlaceholder={t.searchPlaceholder}
+          searchBarText={t.search}
+          onSubmitHref="/searchResult"
+        />
+        <div className="aspect-w-5 aspect-h-2 md:aspect-h-1 xl:ml-3 w-full min-w-max mr-break-out">
+          <Image
+            src={
+              'https://www.canada.ca/content/dam/decd-endc/images/clear-lake-snowy-mountain.png'
```
DaceyTom2, line 34: Can we pass Image src in as a prop so we can reuse components and add tests for the the component?
Also, we might want to consider passing in a component in the future instead of using SearchCard directly for reuse

DaceyTom2 (changes requested):

PR #106, opened 2021-11-10
Missing package

DaceyTom2:
### Description

Updated package to include prop-types as it is missing and another potential cause for the 500 error on Dev


nibivi77 (approved):
Works for me! :)

PR #107, opened 2021-11-15
Content Security Policy (CSP) and Strict-Transport-Security (STS) Headers

TishamIslam:
### Description
Added CSP and STS headers to all pages for our application.

List of proposed changes:
- Added CSP header
- Added SRS header
- other headers that may be required can be added to the header function in `next.config.js`


nibivi77 (approved):
Thank you for including the link for more information. Enables us to understand how this works :)

PR #108, opened 2021-11-15
Dcwc 434 banner

shewood:
### Description
Added the phase banner that is in the Alpha site to our site. It is in the Layout so it should appear on every page ( that was my assumption anyway and can be changed if needed)

List of proposed changes:
Added PhaseBanner component so that it will appear at the top of the pages. 
This is a copy of what Alpha has on their site but some parts that we are not using were removed and it's styled to fit our site.


components/atoms/Card.js
```diff
@@ -11,9 +11,12 @@ export const Card = ({
...
-      <p className="my-5">{text}</p>
+      <div className="">
+        <h3 className="font-bold font-display text-xl">{title}</h3>
+        <p className="">{tag}</p>
+        <p className="my-5">{text}</p>
+      </div>
+
       <ActionButton
         id={btnId}
         className="font-display text-xl underline text-white bg-gray-500 rounded mx-10 md:mx-0 md:px-4 py-2 flex justify-center mx-auto"
```
nibivi77, line 22: `mx-10` and `mx-auto` are being applied. Remove the one we do not use, please :)

components/atoms/Card.js
```diff
@@ -11,9 +11,12 @@ export const Card = ({
 }) => {
   return (
     <div className="flex flex-col place-content-between w-full shadow-card p-6">
-      <h3 className="font-bold font-display text-xl">{title}</h3>
-      <p className="">{tag}</p>
-      <p className="my-5">{text}</p>
+      <div className="">
```
nibivi77, line 14: Do we need this `div`? If we do, can we removed the `className` keyword if we are not using it?

nibivi77 (approved):
Approved! Check the small comment I left :) Good work !

PR #110, opened 2021-11-17
Benefits page

MarcoGoC:
### Description
  based on code already there by Sheila, adding dynamic pages loaded from AEM 
 
List of proposed changes:
- removed the context and fixed the tests
- fixed test for Card and CardList that are using props not coded 

- Download and try 
-


nibivi77 (approved):
This is great thank you Marco.

The `getData.js` will change in Zain's PR.

PR #116, opened 2021-11-22
Banner names and more age options

TishamIslam:
### Description
List of proposed changes:

- Test site banner has text in it now
- Age options on the search filters are now a bit more reasonable


nibivi77 (approved):
That works for me! :)

PR #117, opened 2021-11-22
stopped act wrap error from occurring

TishamIslam:
### Description
As it turns out, [react-testing-library's helpers are all wrapped in act()](https://reactjs.org/docs/testing-recipes.html#act) so the error given by the console when we run tests doesn't make much sense. Mocking Next's Link component fixed the issue


shewood (approved):
Looks great Tisham.

PR #118, opened 2021-11-22
Dcwc 677 top tasks

nibivi77:
### Description

Fetching top tasks from AEM

### Additional Notes

- Added a call for the dictionary in AEM. That is where all titles and other miscellaneous things will reside. There are different ways to implement it so please be constructively judgmental.


shewood (approved):
This looks really good to me. Only thing would be to pull the Alpha out of the path so that it can be updated more easily.

PR #119, opened 2021-11-23
DCWC-858

zeddotes:
- I created a `getPage` method in the AEMService class which uses the same cached method (`getFragment`) in the background to cache page fragments. `getPage` also normalizes the data that is returned from AEM so that consumption of it is easier and predictable.
- I have upgraded the `home` and `search` (previously called `searchResult`...matched with page in AEM) pages to fetch data of the pages content in `getStaticProps` via the `aemServiceInstance`
- The `rewrites` in `next.config.js` is an `async` function that generates all routes available for the app during build time. In order to get all the URLs of the routes the app supports from the URL content stored in AEM, I have leveraged the same `AEMService`*.
- In order to support all this work, I have created an `aemPages` constants file that can be used to declare all internal page IDs (ie. how the app refers to different pages in AEM). These `consts` are referred to in the `aemPages` file that maps each page ID of ours to a path in AEM for its contents (relying on AEM paths for each identification seems strange).


nibivi77 (commented):
This is phenomenal work Zain honeslty. Well done.

Quick question: Would you know what causes this error when you access http://localhost:3000/fr/accueil  directly.

It also makes our `[id].js` crash when trying to acces a specific benefit/service.

image

nibivi77 (approved):
My previous comment has nothing to do with PR. Sorry about that!

zeddotes:
> This is phenomenal work Zain honeslty. Well done.
> 
> Quick question: Would you know what causes this error when you access http://localhost:3000/fr/accueil directly.
> 
> It also makes our `[id].js` crash when trying to acces a specific benefit/service.
> 

Hey @nibivi77, thanks for reviewing this. The issue happens when trying to navigate to /fr/benefits/* page, something that should be discussed with Isabelle and the rest of the team. We have to figure out how the frontend gets parent page names (like /benefits/*) and makes them work with child pages like "oas". This PR isn't meant to fix this issue.

PR #121, opened 2021-11-23
Getting top tasks from AEM

nibivi77:
### Description

Fetching top tasks from AEM

### Additional Notes

- Added a call for the dictionary in AEM. That is where all titles and other miscellaneous things will reside. There are different ways to implement it so please be constructively judgmental.

### DIFF
--- a/__tests__/pages/home.test.js
+++ b/__tests__/pages/home.test.js
@@ -83,6 +83,28 @@ describe('Home page', () => {
     },
   ]
 
+  const topTasks = [
+    {
+      properties: {
+        elements: {
+          scTitleEn: { title: 'titleEn' },
+          scTitleFr: { title: 'titleFr' },
+          scLinkURLEn: { value: '/' },
+          scLinkURLFr: { value: '/' },
+        },
+      },
+    },
+  ]
+
+  const topTaskTitle = {
+    properties: {
+      elements: {
+        scLabelFr: { value: 'test' },
+        scLabelEn: { value: 'test' },
+      },
+    },
+  }
+
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


__tests__/pages/home.test.js
```diff
@@ -83,6 +83,28 @@ describe('Home page', () => {
     },
   ]
 
+  const topTasks = [
+    {
+      properties: {
```
MarcoGoC, line 88: needs to be updated , alpha has 'benefit' instead of 'properties'

pages/home.js
```diff
@@ -13,7 +13,27 @@ import fr from '../locales/fr'
 
 export default function Home(props) {
   const t = props.locale === 'en' ? en : fr
+  const topTasks = () => {
+    return props.topTasks?.map((task) => {
+      try {
+        let title = task.properties.elements.scTitleEn.title
```
MarcoGoC, line 19: same as with the test 'benefit' instead of properties once alpha is working

pages/home.js
```diff
@@ -112,12 +120,38 @@ export async function getStaticProps({ locale }) {
     benefits = AEMbenefits.data.entities
   }
 
+  // Get list of top tasks
+  let topTasksReturned = await aemService.getFragment(
+    'alpha/components/top-tasks.json'
```
MarcoGoC, line 125: base api variable has alpha in it

nibivi77: I need to wait for your PR to be merged before I can remove the alpha or else the benefits inside of CardList.js crashes

DaceyTom2 (approved):

PR #124, opened 2021-11-24
added image priority to imagebox component

shewood:
### Description
Added a bool property for priority to the next/image so that it preloads the image if that image is above the fold. 

more info here. 
https://nextjs.org/docs/api-reference/next/image#priority


MarcoGoC (approved):
no issues here, but why is it an organism??? 
 btw I totally disagree with the classification

PR #129, opened 2021-11-25
Modified comments and formatting

TishamIslam:
### Description
Removed comments that seemed to have no use within the components. I left the comments in the Header.js file as it seemed like we'd be adding breadcrumbs some time in the future.

List of proposed changes:

- removed comments
- edited css class formatting in index.js

### DIFF
--- a/components/molecules/FeatureBlock.js
+++ b/components/molecules/FeatureBlock.js
@@ -14,6 +14,8 @@ export default function FeatureBlock(props) {
       >
         <div className="aspect-w-5 aspect-h-2 sm:aspect-h-1 xl:ml-3 w-full min-w-max mr-break-out">
           <Image
+            // src="/autumn-leaves-woman-hands.png"
+            // hardcoded for now, but should be defined by AEM
             src={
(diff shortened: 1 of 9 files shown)


components/molecules/FeatureBlock.js
```diff
@@ -12,12 +12,8 @@ export default function FeatureBlock(props) {
...
         data-testid="featureBlock"
       >
-        {/* <div className="bg-splash-page bg-cover bg-center w-full py-32  ">
-  
-      </div> */}
         <div className="aspect-w-5 aspect-h-2 sm:aspect-h-1 xl:ml-3 w-full min-w-max mr-break-out">
           <Image
-            // src="/autumn-leaves-woman-hands.png"
             src={
               'https://www.canada.ca/content/dam/decd-endc/images/autumn_leaves_woman_hands.png'
```
MarcoGoC, line 18: the image should be defined in AEM not hardcoded, maybe just leave as it was until this gets decided

TishamIslam: Ok I'll re-add the note and also add that this should be acquired from AEM.

shewood (commented):
Looks like this needs to be merged with dev.

PR #130, opened 2021-11-26
DCWC945-updated e2e tests and a11y

shewood:
### Description
Added some e2e tests for the splash page.
added a11y tests for the splash and home page

List of proposed changes:
Added a role for the banner
the splash page was missing a title
-
### Additional Notes
The click to show the home page is flaky and needs some work. It seems to take some time to display the page.


DaceyTom2 (approved):
Looks good, We will want to touch up the head later to be bilingual and such

PR #131, opened 2021-12-02
Fixed /home error by commenting out AEM scripts

TishamIslam:
### Description

List of proposed changes:

-
-


shewood (approved):
I will approve this for now to let the page work as expected. Tom has already mentioned that he will look into why this is causing the issue in the first place.

PR #135, opened 2021-12-07
DCWC-797

zeddotes:
- Updated `Meta` component to accept a `page` object which should come from AEM.
- Splash page is an exception so the "page" details are passed in statically as opposed to fetching from AEM. Also, there's no language switch functionality on the page itself so EN/FR content will be the same in the `page` object.
- Cleaned up `/benefits/[id]` logic for getting benefit data in `getStaticProps` -- using generic `getPage` method to get the details in a normalized way
- Updated tests

### DIFF
--- a/__tests__/pages/home.test.js
+++ b/__tests__/pages/home.test.js
@@ -32,6 +32,23 @@ jest.mock(
 
 expect.extend(toHaveNoViolations)
 
+const aemPage = {
+  title: {
+    en: 'Home (en)',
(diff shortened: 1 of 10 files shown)


nibivi77 (approved):
This looks good to me. There is a conflict with dev that needs to be looked at first. Thanks Zain!

DaceyTom2 (changes requested):
Refer to https://github.com/DTS-STN/sc-digital-centre/pull/147

PR #136, opened 2021-12-07
Dcwc 988 splash

shewood:
### Description

List of proposed changes:
The purpose was to fix the alignment of the text in the buttons. 
used the action button (as it wasn't created when this page was done) 
Used next/image for the graphic as there is a warning about it in the console,  although it was more of an exercise and example as there isn't much of a benefit from using it in this particular case. 
Updated a some of the tests to add in the home page. 

Make sure that the graphic and the buttons look ok. The active state now looks like Alpha site. 

### Additional Notes
There is some discussion about whether the action button can be split into a link button and an action button and we can still do that.

### DIFF
--- a/cypress/integration/home.spec.js
+++ b/cypress/integration/home.spec.js
@@ -2,15 +2,15 @@
 
 describe('home page loads', () => {
     beforeEach(() => {
-      cy.visit('/')
-      // cy.injectAxe();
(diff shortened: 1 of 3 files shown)


DaceyTom2 (approved):
Looks good, I'd like to see Link and Button separated out of action button

PR #139, opened 2021-12-13
Encoding query string

nibivi77:
### Description

Added encoding to the query string with cypress tests.


yeshwanthjs (approved):

DaceyTom2 (changes requested):
Looks reasonable!

DaceyTom2 (approved):
Miss clicked approve

PR #143, opened 2021-12-16
Apply layout to all pages

DaceyTom2:
### Description
I propose we apply layout to all pages. This allows us to tackle things globally in our app even with exceptions to the layout.
By default header/footer exist but can be toggled off for pages that do not require them.

List of proposed changes:

- Apply layout to Index.js
- Header/Footer in layout can be toggled

### Additional Notes
It may be unnecessary to apply changes to Layout since we have the option to use _document.js as well

### DIFF
--- a/components/organisms/Layout.js
+++ b/components/organisms/Layout.js
@@ -19,6 +19,8 @@ export default function Layout({
   phase,
   bannerText,
   aemPage,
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Good work!

PR #144, opened 2021-12-21
Aem revamp

DaceyTom2:
### Description
Moved AEM calls out of page static props

List of proposed changes:

- Decoupled AEM from the app
- AEM calls are mostly async to help with loading times in development

Ensure app runs without changes or errors


lib/pageContent.js
```diff
@@ -46,7 +46,10 @@ export async function getHomePageContent(locale) {
     miscellaneousRes.data.entities.forEach((item) => {
       // Extracting Top Task component content (Title and whatever else we add in AEM later on)
       if (item.properties.elements.scId.value === 'TOP-TASKS') {
-        topTasksTitle = item.properties.elements.scLabelEn.value
+        topTasksTitle =
+          locale === 'en'
+            ? item.properties.elements.scLabelEn.value
+            : item.properties.elements.scLabelFr.value
```
MarcoGoC, line 52: Look at Zain way of getting these values. that being said I don't understand the purpose of the changes. it s a lot cleaner as is right now.

DaceyTom2: I think you are referring to the wait Zain had AEMObj.[Locale].value?
If so that isn't possible due to the structure in AEM.

The purpose of the PR is to decouple AEM from the app. The advantage is when going to GraphQL we can just change the requests and mappings in pageContent. Also, the components will be reusable and not locked to specific AEM model.

shewood (commented):
I think it looks good but maybe call pageContent cmsContent ?

shewood (approved):
Looks good

PR #146, opened 2021-12-22
Dc 1008 badges

shewood:
### Description
Added badges for e2e and unit tests passing or failing. 
Added the links for the badges for main and dev

I am not sure the main ones will work. ??


DaceyTom2 (approved):
Looks like it should work. Not sure where the report name comes from.

PR #149, opened 2021-12-24
Dcwc 1105 clean up pages

TishamIslam:
### Description

List of proposed changes:

- Extracted components such that the page code would be more readable


DaceyTom2 (changes requested):
Looks really good! Just need tests for SplashLanguageSelect and MostRequestedPages

PR #152, opened 2021-12-29
Dcwc 1109 standardise props

DaceyTom2:
### Description
Standardising props

List of proposed changes:

- Cleaning up specified props to only be defined in proptypes

Ensure app still runs smoothly

### DIFF
--- a/components/atoms/ActionButton.js
+++ b/components/atoms/ActionButton.js
@@ -4,47 +4,38 @@ import Link from 'next/link'
 /**
  * Button component
  */
-export function ActionButton({
-  href,
-  className,
-  onClick,
-  type,
-  lang,
-  id,
-  text,
-  children,
-  ...rest
-}) {
+export function ActionButton(props) {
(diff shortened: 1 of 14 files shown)


components/atoms/ActionButton.js
```diff
@@ -4,47 +4,38 @@ import Link from 'next/link'
...
-  onClick,
-  type,
-  lang,
-  id,
-  text,
-  children,
-  ...rest
-}) {
+export function ActionButton(props) {
+  var { ...rest } = props
```
nibivi77, line 8: Use `const` if possible, please. Otherwise, use `let` :)  Gotta break those old habits ;)

nibivi77 (changes requested):

components/atoms/ActionButton.js
DaceyTom2: Why am I like this

nibivi77 (approved):

PR #160, opened 2022-01-07
shortened all bilingual mappings

TishamIslam:
### Description
Shortened bilingual mappings by using `object[keyName + locale].value` instead of the old 3-line ternary expressions.

List of proposed changes:
- Shortened bilingual mappings


nibivi77 (approved):
My name is Benoit Cloutier, and I approve this message!

P.S: Most of this will have to change I assume when and if GraphQL will be used.

PR #161, opened 2022-01-07
Dcwc xxx i18n fixv3 - Part 2/2

MarcoGoC:
very small list of changes but it also contains part 1/2 

List of proposed changes:
- fix the language toggle removing unnecessary id query 
- fix the benefits href address 

- see deployment branch


nibivi77 (approved):
Thank you for this marco! The bug was so annoying. i am just curious as to why when we run this branch locally, it's slower than dev, even if we remove a call to AEM here. Nonetheless, looks good!

PR #162, opened 2022-01-10
updated yml file for tests

shewood:
### Description
Commented out a line in the yml for action and changed another line to hopefully work for both main and dev if there is no comment already.


MarcoGoC (approved):
I still don't know why

PR #180, opened 2022-01-28
circle progress bar component

nabil386:
### Description
circle progress bar component
List of proposed changes:

- CircleProgressBar.js
-


amirpourpanahi (approved):

styles/stepper.css
```diff
@@ -104,3 +104,36 @@
         padding-left: 1rem;
     }
 }
+
+.circle-progress-indicator {
```
arsanig, line 108: This shouldn't be a part of the `stepper.css` file and would be more maintainable if it's in its own file.

styles/stepper.css
```diff
@@ -104,3 +104,36 @@
...
+    font-family: Lato;
+    font-style: normal;
+    font-weight: bold;
+    font-size: 20px;
+    line-height: 30px;
+    color: #259F46;
+    letter-spacing: 0.1px;
+}
+
+.RCP {
```
arsanig, line 130: What's RCP? Can this `classname` be made more descriptive so it's easier to identify what it's targeting in the future?

styles/stepper.css
```diff
@@ -104,3 +104,36 @@
...
+    width: 50px;
+    margin-left: 0px;
+
+}
+
+svg {
```
arsanig, line 136: Targeting all `<svg>` attributes is not a good practice. Tailwind prefers an inline svgs with a `class=` attribute to modify svgs

arsanig (changes requested):
Please see feedback below

styles/stepper.css
nabil386: it has its own css now

styles/stepper.css
nabil386: .RCP {}
.RCP__track {}
.RCP__progress {}
.RCP__pointer {}
are styles of react-customizable-progressbar component and .RCP alone was a leftover and it's not needed. removed it

styles/stepper.css
nabil386: done

styles/stepper.css
```diff
@@ -104,3 +104,4 @@
         padding-left: 1rem;
     }
 }
+
```
arsanig, line 107: How come this is a part of the PR? Don't think this needs to be here if there isn't any changes to this file

nabil386: file got reverted

styles/stepper.css
```diff
@@ -9,7 +9,7 @@
     margin-bottom: 16px;
     margin-left: 10px;
 }
-  
+
```
arsanig, line 12: Still not sure if this is needed as part of the PR if there are no changes to this file

components/atoms/CircleProgressBar.js
```diff
@@ -0,0 +1,32 @@
...
+
+export class CircleProgressBar extends Component {
+  render() {
+    return (
+      <div className="absolute inset-x-6 top-9 h-16">
+        <ProgressBar
+          progress={this.props.progress}
+          steps={this.props.steps}
+          radius={28}
+          strokeColor={'#259F46'}
```
arsanig, line 14: Can you add this colour to the `tailwind.config.js`?

arsanig (changes requested):
Amazingly clean compared to the custom .css sheet we had before 👏

Other fix - the `package-lock` is built with an old npm, can you update your npm locally and then push those changes?

1. `npm i -g npm@8.4.0`
2. `npm audit fix`
3. `npm i`

nabil386:
what do we do with that color if we added to tailwind ?

arsanig (approved):

PR #182, opened 2022-01-31
Creating ActiveBenefitDetails component

amirpourpanahi:
### Description

List of proposed changes:

- Creating ActiveBenefitDetails component that works for all active benefits
- Adding CPP active Card


components/molecules/PendingBenefitDetails.jsx
```diff
@@ -0,0 +1,64 @@
...
+import en from '../../locales/en'
+import fr from '../../locales/fr'
+
+const PendingBenefitCard = (props) => {
+  const t = props.locale === 'en' ? en : fr
+
+  return (
+    <div className="grid grid-cols-1 divide-y divide-grey-500 font-display my-4">
+      <section>
+        <div className="sm:flex sm:justify-around sm:grid sm:grid-cols-7 sm:gap-4 sm:mx-auto sm:mt-4">
```
arsanig, line 13: There's both flex and grid assigned on this and I don't think you need the flex

components/molecules/PendingBenefitDetails.jsx
```diff
@@ -0,0 +1,64 @@
...
+          <div
+            id="claimUpdates"
+            className="col-span-2 sm:col-span-3 lg:col-span-2 sm:w-60 ml-7 mb-6 sm:mb-3"
```
arsanig, line 16: I think this broke the mobile work that was completed as part of PR #181 

image

components/molecules/PendingBenefitDetails.jsx
```diff
@@ -0,0 +1,64 @@
...
+            <VerticalStepper
+              benefitStatus={props.benefit.status}
+              lastUpdates={props.benefit.lastUpdates}
+              locale={props.locale}
+            />
+          </div>
+
+          <div
+            id="details"
+            className="col-span-5 sm:col-span-4 lg:col-span-5 sm:grid mr-5"
```
arsanig, line 27: `className="col-span-5 sm:col-span-4 lg:col-span-5 sm:grid mr-5"`

Don't need col-span-5, twice here. Something like below will accomplish the same thing:
`className="col-span-4 mr-5 sm:grid sm:col-span-5"`

components/molecules/PendingBenefitDetails.jsx
```diff
@@ -0,0 +1,64 @@
...
+          >
+            <div id="application_details" className="ml-5 mb-5 sm:mb-0">
+              <p className="font-bold text-xl">{t.applicationDetails}</p>
+              <p className="font-normal text-base">
+                {props.benefit.statusDescription}
+              </p>
+            </div>
+            <div id="application_description" className="ml-5 mb-5 sm:mb-0">
```
arsanig, line 35: Instead of `ml-5` per item here, just put it on the parent `<div>`

arsanig (changes requested):
Looks good! 👍🏻 Though I think there were a few things missed with the local merged you mentioned you performed, see feedback below.

components/molecules/PendingBenefitDetails.jsx
amirpourpanahi: flex is removed

components/molecules/PendingBenefitDetails.jsx
amirpourpanahi: Do you mean the Left Margin?
I just changed the name of this file. I worked on Active card.

components/molecules/PendingBenefitDetails.jsx
amirpourpanahi: mobile-screens
https://tailwindcss.com/docs/responsive-design

components/molecules/PendingBenefitDetails.jsx
amirpourpanahi: Done

components/molecules/PendingBenefitDetails.jsx
arsanig: This file is different than its previous `PendingBenefitCard.jsx` something changed somewhere.

nabil386 (approved):

components/molecules/PendingBenefitDetails.jsx
arsanig: Ok sounds good

PR #204, opened 2022-02-16
Scroll view more for expanding or collapsing cards

arsanig:
### Description
List of proposed changes:
- Added a function that is called to scroll to top of card when view less is clicked or top of tasks when view more is clicked

- Updated BenefitCard tests to use the correct constants 

### Additional Notes
- Doesn't scroll properly on the last card as there isn't a component below the last card that allows for enough height for the animation to occur


components/molecules/BenefitCard.test.js
```diff
@@ -2,55 +2,15 @@ import { render, screen } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 import { axe, toHaveNoViolations } from 'jest-axe'
 import BenefitCard from './BenefitCard'
-
-const activeCanadaPensionPlan = {
```
nabil386: can we leave the test data separate ?

__tests__/pages/__snapshots__/dashboard.test.js.snap
```diff
@@ -2,7 +2,9 @@
 
 exports[`Dashboard renders Dashboard 1`] = `
 <div>
-  <div>
+  <div
```
nabil386: this might need to be generated

nabil386 (approved):

amirpourpanahi (approved):

PR #205, opened 2022-02-16
add tasks text and icons for ei card

nabil386:
### Description

List of proposed changes:

- add text and icons to EI card
-


amirpourpanahi (approved):

components/molecules/BenefitCard.test.js
```diff
@@ -47,10 +47,64 @@ const activeCanadaPensionPlan = {
   ],
 }
 
+const activeCppTasks = [
```
arsanig, line 50: This can just be imported from constants instead of adding it here too. I already updated this test as part of #204

arsanig (changes requested):

PR #224, opened 2022-03-01
Install Design System

amirpourpanahi:
### Description

List of proposed changes:

- Installing design system package
-


DaceyTom2 (approved):
Not a huge deal but we are running different versions of npm. package-lock is changed to v1 instead of v2. Whoever is below npm v7 should update.

Also, LG2M appears to be hanging up

PR #241, opened 2022-03-21
Added design system header and footer with tests

arsanig:
### Description
List of proposed changes:
- Created a component that pulls in the Design System's Header
- Created a component that pulls in the Design System's Footer

- Proper rendering of the component (more tests will have to be added down the road when the menu's actually link)
- No a11y violations
### Additional Notes
- Dashboard page should be cleaned up as it's already starting to get cluttered, maybe a `<BenefitCards />` component holding all the benefit cards?


amirpourpanahi (approved):

nabil386 (approved):

DaceyTom2 (approved):
Looks reasonable. The header and footer can be added to a layout in time. Being directly on the page for now works

PR #257, opened 2022-04-01
Feature/dcwc 1184 add error pages

nabil386:
### Description

List of proposed changes:

- add error page 400 http response code
- add error page for 503 http response code

### DIFF
--- a/__tests__/pages/dashboard.test.js
+++ b/__tests__/pages/dashboard.test.js
@@ -7,7 +7,7 @@ import Dashboard from '../../pages/dashboard'
 expect.extend(toHaveNoViolations)
 
 describe('Dashboard', () => {
-  jest.setTimeout(25)
+  jest.setTimeout(10000)
   it('renders Dashboard', () => {
     const { container } = render(<Dashboard />)
     expect(container).toMatchSnapshot()


pages/400.js
```diff
@@ -0,0 +1,14 @@
+import Link from 'next/link'
+
+export default function Error400() {
+  return (
+    <div>
+      <h1>400 Error - Bad request error occurred</h1>
+      <Link href="/dashboard">
+        <a className="text-sm hover:text-bright-blue-solid visited:text-purple-solid underline">
+          Return to dashboard main page
```
amirpourpanahi, line 9: I think it's better to move it to locales/en.js file

pages/503.js
```diff
@@ -0,0 +1,3 @@
+export default function Error503() {
+  return <h1>503 - Service unavailable error occurred</h1>
```
amirpourpanahi, line 2: Same here!

pages/400.js
nabil386: done in b18b6c06469fc4da56ab4cde1926e2e1275d6e73

pages/503.js
nabil386: done in b18b6c06469fc4da56ab4cde1926e2e1275d6e73

s-laugh (commented):
I don't think these pages will work the way you expect them too without proper routing in place... I'd suggest looking at https://nextjs.org/docs/advanced-features/custom-error-page

DaceyTom2:
The error pages seem to work fine without any custom routing or custom error page. Other than the minor A11y issue the PR looks good

DaceyTom2 (approved):

amirpourpanahi (approved):

PR #258, opened 2022-04-01
Left align the icons

amirpourpanahi:
### Description

List of proposed changes:

- Left align the icons
-

### DIFF
--- a/components/molecules/BenefitTasks.js
+++ b/components/molecules/BenefitTasks.js
@@ -21,7 +21,7 @@ export default function BenefitTasks(props) {
               <a href={value.taskLink}>
                 <img
                   src={value.taskIcon}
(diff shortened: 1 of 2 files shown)


components/molecules/BenefitTasks.js
```diff
@@ -21,7 +21,7 @@ export default function BenefitTasks(props) {
               <a href={value.taskLink}>
                 <img
                   src={value.taskIcon}
-                  className="font-normal pb-3 px-8"
+                  className="font-normal pb-3 px-4"
```
s-laugh, line 24: the `p` tag is using a `px-5` should this not be the same to match it?

amirpourpanahi: Fixed

nabil386 (approved):

PR #265, opened 2022-04-14
Import DS LayoutContainer

amirpourpanahi:
### Description

List of proposed changes:

- Import Design System LayoutContainer component
-


s-laugh (changes requested):
This seems to introduce a styling bug for when the screen size is just under ~1005px wide until ~975px wide, where the program name in the cards shrinks before the card does.
image

amirpourpanahi:
> This seems to introduce a styling bug for when the screen size is just under ~1005px wide until ~975px wide, where the program name in the cards shrinks before the card does; and also with the right side padding of the coloured status. image


I fixed the issue of padding for title and colored status. 
Since the size of card is changing  a little in your mentioned wides, if the title don't shrink, some long tiles go on the vertical bar.

s-laugh (approved):
Great.
As we discussed, it does change a little; but the cards are now at least uniform in a given size (users aren't expected to changes sizes)

nabil386 (approved):

PR #269, opened 2022-04-19
DCWC-1557 Bold lettering

CurtisUnderwood:
### Description

List of proposed changes:

- Fixed bold lettering issue on Next Payment Date Text


nabil386 (approved):

shewood (approved):
This looks good but should it also have text-lg so that the font size is the same as the one in black in the card above/below?

PR #278, opened 2022-04-20
DCWC-1588 Add the header to the error pages

CurtisUnderwood:
### Description

List of proposed changes:

- Added header to error pages manually
- Fixed small issue with previous PR, changed text to text-lg

### Additional Notes
- Error page components have pre-existing layout containers built-in, will need to consider this if we build a custom layout that includes the header


s-laugh (approved):
Looks good

PR #281, opened 2022-04-20
Feature/dcwc 1530 dcwc 1531 cpp ei api

nabil386:
### Description

List of proposed changes:

- connect active cpp API to active cpp card
- connext active ei API to active ei card


amirpourpanahi (approved):

s-laugh (changes requested):
I'm getting quite lost in the logic here; would be great to see less logic in the components and just mapping to a state. It would help with the scalability as all these cards will be eventually populated by the API

components/organisms/DashboardUtils.js
```diff
@@ -0,0 +1,37 @@
...
+export function getBenefitCode(benefitCode) {
+  let benefit = BENEFIT_CODES.filter(function (item) {
+    if (item.code == benefitCode) {
+      return item
+    }
+  })
+  return benefit
+}
+
+export function getBenefitType(benefitType) {
```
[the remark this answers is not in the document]
nabil386: the only way to make this more generic is to create code / value. Now the codes are broken down to tables program codes, benefit codes, etc..

.github/workflows/test-and-lint.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         env:
           NEXT_CONTENT_API: ${{ secrets.NEXT_CONTENT_API }}
           NEXT_PUBLIC_ADOBE_ANALYTICS_URL: ${{ secrets.NEXT_PUBLIC_ADOBE_ANALYTICS_URL}}
+          OCP_APIM_SUBSCRIPTION_KEY: ${{ secrets.OCP_APIM_SUBSCRIPTION_KEY }}
+          CPP_ACTIVE_BENEFIT_URL: ${{ secrets.CPP_ACTIVE_BENEFIT_URL }}
+          EI_ACTIVE_BENEFIT_URL: ${{ secrets.EI_ACTIVE_BENEFIT_URL }}
+          APP_URL: "https://sc-digital-centre-dashboard.bdm-dev.dts-stn.com:443"
```
s-laugh, line 45: Again here, a dynamic URL is likely needed.

.teamcity/settings.kts
```diff
@@ -79,17 +79,18 @@ object Build_Release: BuildType({
...
         param("env.NEXT_CONTENT_API", "%vault:dts-secrets-dev/data/digitalCentre!/NEXT_CONTENT_API%")
         param("env.NEXT_PUBLIC_ADOBE_ANALYTICS_URL", "%vault:dts-secrets-dev/data/digitalCentre!/NEXT_PUBLIC_ADOBE_ANALYTICS_URL%")
-        param("env.OCP_APIM_SUBSCRIPTION_KEY", "%vault:dts-secrets-dev/data/digitalCentre!/OCP_APIM_SUBSCRIPTION_KEY%")
-        param("env.CPP_ACTIVE_BENEFIT_URL", "%vault:dts-secrets-dev/data/digitalCentre!/CPP_ACTIVE_BENEFIT_URL%")
-        param("env.EI_ACTIVE_BENEFIT_URL", "%vault:dts-secrets-dev/data/digitalCentre!/EI_ACTIVE_BENEFIT_URL%")
-        param("env.TARGET", "dev")
+        param("env.OCP_APIM_SUBSCRIPTION_KEY", "%vault:dts-secrets-dev/digitalCentre!/OCP_APIM_SUBSCRIPTION_KEY%")
+        param("env.CPP_ACTIVE_BENEFIT_URL", "%vault:dts-secrets-dev/digitalCentre!/CPP_ACTIVE_BENEFIT_URL%")
+        param("env.EI_ACTIVE_BENEFIT_URL", "%vault:dts-secrets-dev/digitalCentre!/EI_ACTIVE_BENEFIT_URL%")
+        param("env.APP_URL", "https://sc-digital-centre-dashboard.bdm-dev.dts-stn.com:443")
```
s-laugh, line 85: It seems there are a few of these `app_urls` that don't match the branch/target in this file.

s-laugh (changes requested):
This is looking good, could you also update the .env.example file?

.github/workflows/e2e-test.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         env:
           NEXT_CONTENT_API: ${{ secrets.NEXT_CONTENT_API }}
           NEXT_PUBLIC_ADOBE_ANALYTICS_URL: ${{ secrets.NEXT_PUBLIC_ADOBE_ANALYTICS_URL}}
+          OCP_APIM_SUBSCRIPTION_KEY: ${{ secrets.OCP_APIM_SUBSCRIPTION_KEY }}
+          CPP_ACTIVE_BENEFIT_URL: ${{ secrets.CPP_ACTIVE_BENEFIT_URL }}
+          EI_ACTIVE_BENEFIT_URL: ${{ secrets.EI_ACTIVE_BENEFIT_URL }}
+          APP_URL: "https://sc-digital-centre-dashboard.bdm-dev.dts-stn.com:443"
```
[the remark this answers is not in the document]
s-laugh: I'm thinking maybe it shouldn't be an env variable at all but instead be using a relative path in the api call https://github.com/DTS-STN/sc-digital-centre/pull/281/files#diff-1aca893454c20bb454daf457d86f32ddfd829f21a6cd96f0f2310a29284f327cR198

.teamcity/settings.kts
nabil386: same thing like the above discussion

s-laugh: This one is different - while it would be nice to be dynamic, it at least needs to match the build it's targeting; here the url should be `https://sc-digital-centre-release.bdm-dev.dts-stn.com:443` if it has to be hard coded.

.github/workflows/e2e-test.yml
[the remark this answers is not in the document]
s-laugh: As disscussed, we can remove the APP_URL env varriable and from the dashboard we can call interop API directly with `const res = await fetch(process.env.EI_ACTIVE_BENEFIT_URL, {`... ext

s-laugh (approved):

PR #287, opened 2022-04-22
DCWC-1614 Build Self Employment Card

CurtisUnderwood:
### Description

List of proposed changes:

- Added Self Employment Benefits Card
- Created a new Header BenefitHeaderActiveAgreement for this new type of card header
- Added small padding to the card header that appears in the Figma (this may need to be applied to the other headers)

### DIFF
--- a/components/atoms/BenefitCardHeaderActiveAgreement.test.js
+++ b/components/atoms/BenefitCardHeaderActiveAgreement.test.js
@@ -1,0 +1,39 @@
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import { axe, toHaveNoViolations } from 'jest-axe'
+import BenefitCardHeaderActiveAgreement from './BenefitCardHeaderActiveAgreement'
+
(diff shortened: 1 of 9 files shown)


components/atoms/BenefitCardHeaderActiveAgreement.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+    expect(titleText).toBeInTheDocument()
+    expect(benefitStatusProgress).toBeInTheDocument()
+    expect(transactionDate).toBeInTheDocument()
+  })
+
+  it('has no a11y violations', async () => {
+    const { container } = render(
+      <BenefitCardHeaderActiveAgreement benefit={ACTIVE_SEB} locale={'en'} />
+    )
```
s-laugh, line 35: If you pull the render out of both `it()` functions, but the first thing in the `describe()` - we only need to render it once, speeding up the tests. (this only works as you're sending the same props in both renders)

s-laugh (approved):
Great work!

PR #296, opened 2022-04-25
deleting old patch and updating the env vars (and fixing the vault path)

P0NDER0SA:
Please review the additional env variables to make sure they are good.  Paths have been fixed.


s-laugh (approved):
looks good

PR #297, opened 2022-04-25
CPP Pending - Adding task group

amirpourpanahi:
### Description

List of proposed changes:

- Adding task group for CPP pending card


s-laugh:
You might want to double check with Breanna, I believe a design decision was made at the last design review to not group tasks when they are less than 5 - so this change might have to get un-done.

nabil386 (approved):

PR #298, opened 2022-04-25
Adding Profile Information Section.

nibivi77:
### Description

- Task to be created.
- Layout will come with a later PR
- Figma design: https://www.figma.com/file/wLUT8UWdL6MjSD8j8vHw0t/Dashboard---Design-File?node-id=7373%3A149000

### DIFF
--- a/components/atoms/ProfileInfoSection.js
+++ b/components/atoms/ProfileInfoSection.js
@@ -8,17 +8,15 @@ export default function ProfileInfoSection(props) {
         {props.info.map((item) => {
           return (
             <div key={item.title}>
-              <span>
-                <span className="flex justify-between">
-                  <h3 className="font-bold">{item.title}</h3>
-                  <button className="underline text-blue-600 cursor-pointer hover:text-blue-800 visited:text-purple-600">
-                    Edit
-                  </button>
-                </span>
-                {item.fields?.map((field) => (
-                  <p key={field}>{field}</p>
-                ))}
+              <span className="flex justify-between">
+                <h2 className="font-bold">{item.title}</h2>
+                <button className="underline text-blue-600 cursor-pointer hover:text-blue-800 visited:text-purple-600">
+                  Edit
+                </button>
               </span>
+              {item.fields?.map((field) => (
+                <p key={field}>{field}</p>
+              ))}
             </div>
           )
         })}
--- a/components/atoms/ProfileInfoSection.js
+++ b/components/atoms/ProfileInfoSection.js
@@ -1,0 +1,39 @@
+import propTypes from 'prop-types'
+
+export default function ProfileInfoSection(props) {
+  return (
+    <>
+      <h3 className="border-b border-gray-300 font-bold">{props.title}</h3>
+      <div className="grid sm:grid-cols-2 gap-y-8 gap-x-40 py-8">
+        {props.info.map((item) => {
+          return (
+            <div key={item.title}>
+              <span>
+                <span className="flex justify-between">
+                  <h3 className="font-bold">{item.title}</h3>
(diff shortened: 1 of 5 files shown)


components/molecules/ProfileInfo.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe('ProfileInfo', () => {
+  it('renders ProfileInfo', () => {
+    render(<ProfileInfo />)
+    const text = screen.getByText('Ontario')
+    expect(text).toBeInTheDocument()
+  })
+
+  it('has no a11y violations', async () => {
+    const { container } = render(<ProfileInfo />)
```
s-laugh, line 16: SUGESTING: If we move the render outside the `it` functions, into the `describe`, we only need to render the component once during the tests (speeding it up slightly).

components/molecules/ProfileInfo.test.js
```diff
@@ -0,0 +1,20 @@
...
+import '@testing-library/jest-dom/extend-expect'
+import { axe, toHaveNoViolations } from 'jest-axe'
+import ProfileInfo from './ProfileInfo'
+
+expect.extend(toHaveNoViolations)
+
+describe('ProfileInfo', () => {
+  it('renders ProfileInfo', () => {
+    render(<ProfileInfo />)
+    const text = screen.getByText('Ontario')
```
s-laugh, line 11: If you pass the values in as props you can manually set them in the test, ensuring that the test will work independent of the data in the component.

components/atoms/ProfileInfoSection.js
```diff
@@ -0,0 +1,39 @@
...
+
+export default function ProfileInfoSection(props) {
+  return (
+    <>
+      <h3 className="border-b border-gray-300 font-bold">{props.title}</h3>
+      <div className="grid sm:grid-cols-2 gap-y-8 gap-x-40 py-8">
+        {props.info.map((item) => {
+          return (
+            <div key={item.title}>
+              <span>
```
s-laugh, line 11: SUGESTING: this `span` seems redundant here

components/atoms/ProfileInfoSection.js
```diff
@@ -0,0 +1,39 @@
+import propTypes from 'prop-types'
+
+export default function ProfileInfoSection(props) {
+  return (
+    <>
+      <h3 className="border-b border-gray-300 font-bold">{props.title}</h3>
```
s-laugh, line 6: Double check your heading levels, this needs to be an h2 (I also don't see an h1, but this isn't where that would go)
```suggestion
      <h2 className="border-b border-gray-300 font-bold">{props.title}</h2>
```

s-laugh (changes requested):
Looks like a great start, just needs a few tweaks.

pages/profile.js
```diff
@@ -1,14 +1,12 @@
...
+import ProfileInfo from '../components/molecules/ProfileInfo'
 
 export default function Profile(props) {
   return (
     <>
       <DSHeader locale="en" />
-      <LayoutContainer>
-        <div>Profile</div>
-      </LayoutContainer>
+      <ProfileInfo />
```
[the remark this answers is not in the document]
nibivi77: Layoutcontainer splits the page into 12 columns and makes it look squeezed. I am also working with Curtis on this page so the layout will come when both of our work combines together.

components/molecules/ProfileInfo.test.js
nibivi77: Great suggestion but this prevents us from rendering the component differently for future test cases. Since this component will take in props/data, we will most likely render it with different data and therefore, rendering the component differently.

PR #303, opened 2022-04-27
Create Universal Benefit Object

s-laugh:
### Description

List of proposed changes:

- adding some classes to discus with team about their use and mapping to API calls data returns
-


nibivi77 (approved):
Seems to be a good start for now!

PR #305, opened 2022-04-28
Map api generic object

s-laugh:
### Description

List of proposed changes:

- adding logic to map api data to object
-


nibivi77 (approved):
I understand how this works and the idea of it. I think this is a good starting point and prototype to evaluate before we go too far.

PR #306, opened 2022-04-28
using the codes to match data

s-laugh:
### Description

List of proposed changes:

- adding the links to map API result to advertising cards display logic
-


CurtisUnderwood (approved):
Works!

PR #307, opened 2022-04-29
Universal benefit card skeleton

s-laugh:
### Description

List of proposed changes:

- Just adding the skeleton for how these new components will work with each other 
-


components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
+import propTypes from 'prop-types'
+import { SummaryTypes } from '../../objects/UniversalBenefit'
+
+export default function BenefitCardHeaderSummary(props) {
+  const t = props.locale
+  let typesWithLinks = [SummaryTypes.PaymentAmount, SummaryTypes.LatestStatus]
```
nibivi77, line 6: Lets try to use `const` instead of `let` as much as we can :)

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <div>
+      <p>{t[props.summary.type].title}</p>
```
nibivi77, line 10: Maybe this should be a head since it's a title? `<h2>` or something like that

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
...
+      <p className="font-bold">
+        {props.summary.value}
+        <br />
```
nibivi77, line 13: `<br />` should be avoided for a11y issues.

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
...
+        {props.summary.status ?? null}
+      </p>
+      {!typesWithLinks.find((t) => t == props.summary.type) ? null : (
```
nibivi77, line 16: Try to use triple equals when we can please :) Makes the code more robust.

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
...
+        <p>
+          <a href={t[props.summary.type].link}>
```
nibivi77, line 18: When using `<a>` tags that actually navigates to somewhere else in Next.js, we should try to always use the provided Next.JS `<Link>` component.

https://stackoverflow.com/questions/65086108/next-js-link-vs-router-push-vs-a-tag

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <div>
+      <p>{t[props.summary.type].title}</p>
+      <p className="font-bold">
+        {props.summary.value}
+        <br />
+        {props.summary.status ?? null}
+      </p>
+      {!typesWithLinks.find((t) => t == props.summary.type) ? null : (
+        <p>
```
nibivi77, line 17: Not sure if we need this surrounding <p> tag. Could be removed with some styling. If not, no problem!

components/molecules/BenefitCardHeaderSummary.test.js
```diff
@@ -0,0 +1,32 @@
...
+import {
+  CreateBenefitSummary,
+  SummaryTypes,
+} from '../../objects/UniversalBenefit'
+import BenefitCardHeaderSummary from './BenefitCardHeaderSummary'
+
+expect.extend(toHaveNoViolations)
+
+describe('BenefitCardHeaderSummary', () => {
+  let netPay = 30
```
nibivi77, line 14: Same for this whole file. `const` if possible :)

__tests__/pages/dashboard.test.js
```diff
@@ -43,8 +51,19 @@ describe('Dashboard', () => {
     APPLICATION_CARD_EI,
     APPLICATION_CARD_CPP,
   ]
+  let usersBenefits = [
```
nibivi77, line 54: `const` if possible. If not, no problem!

components/molecules/UniversalBenefitCard.js
```diff
@@ -0,0 +1,37 @@
...
+import BenefitCardHeaderSummary from './BenefitCardHeaderSummary'
+
+export default function UniversalBenefitCard(props) {
+  const t = props.locale === 'en' ? en : fr
+
+  return (
+    <div className="benefit-card">
+      {/* Status Bar Component goes here */}
+      <h2>{t[props.benefit.programCode]}</h2>
+      {props.benefit.summaries == null || props.benefit.summaries.length <= 0
```
nibivi77, line 14: When checking for null, I think you can do something like `props.benefit.summaries` and if it's not null, it will default to true.

i think you can remove the `== null`.

components/molecules/UniversalBenefitCard.test.js
```diff
@@ -0,0 +1,34 @@
...
+  CreateBenefitCardObj,
+  ProgramCodes,
+  StatusCodes,
+  TypeCodes,
+} from '../../objects/UniversalBenefit'
+
+expect.extend(toHaveNoViolations)
+
+describe('UniversalBenefitCard', () => {
+  let benefit = CreateBenefitCardObj(
```
nibivi77, line 16: `const` :)

objects/UniversalBenefit.js
```diff
@@ -21,6 +24,32 @@ export const StatusCodes = {
...
+  EIUnknown: 'EIUnknown',
+}
+
+export function CreateBenefitSummary(type, value, status) {
+  let benefitSummary = {
+    type: type, // will define what header text to go with it and any links if applicable
+    value: value, // a date or amount, defined by the type
+  }
+  //only add the status object if it exists
+  if (status != undefined) {
```
nibivi77, line 40: Same as null a null check here. I think you can change this as `if (!status) {`. If not, use triple equals :)

pages/dashboard.js
```diff
@@ -74,6 +75,20 @@ export default function Dashboard(props) {
       <LayoutContainer>
         <div className="col-span-12 mb-8">
           <Greeting locale={props.locale} name="Mary" />
+
+          {/* New Benefit Cards API driven */}
+          {props.usersBenefits == null || props.usersBenefits.length <= 0
```
nibivi77, line 80: null check ;)

nibivi77 (changes requested):
Sorry for my nitpicking here. I figured I'd give my review before merging it. Don't hesitate to ask me anything if you have questions or disagree with anything.

components/molecules/BenefitCardHeaderSummary.js
s-laugh: Hmmm maybe, it would have the most effect on a screen reader, and I'm not sure it's actually desired based on the rest of the card (the task group have titles that are a header). Either way this is something for PO & Design team to talk about with. 
Since it's currently a `p` in the old benefitCard it's best to leave it for now.

s-laugh: nvm I agree it should be a header of some kind for a11y - but I think maybe these "summaries" need to be under their own header - which isn't in the design. Still something to raise with them.

components/molecules/BenefitCardHeaderSummary.js
s-laugh: yes and no.... but pretty sure you're right in this case

components/molecules/BenefitCardHeaderSummary.js
s-laugh: content should always be wrapped in a `p`, and links would be considered content. Buttons and headers would be the exclusion.

components/molecules/UniversalBenefitCard.js
s-laugh: But I'm looking for a true result when it is null; removing it gives me the opposite; so it is needed here.

s-laugh: and we actually want this to be a double equals so that if summaries is null or undefined we get a true.

components/molecules/BenefitCardHeaderSummary.js
s-laugh: and I'm correcting myself again, we can't make them headers outright because they would be repeated on the page

nibivi77: Fair enough! :)

components/molecules/UniversalBenefitCard.js
nibivi77: Sounds good with me.

nibivi77 (approved):
Amazing work Shaun! Thank you!

PR #308, opened 2022-04-29
added status badge component

shewood:
### Description

List of proposed changes:
Moved the status badge to its own component and added it to the card headers. 
The badge part will take the whole top of the card on mobile and then move to the top of the card from sm onwards.


s-laugh (approved):
Looks great! The new component StatusBage should play really nicely with the new cards I'm building.

PR #316, opened 2022-05-04
DCWC-1706 Added universal status bar

CurtisUnderwood:
### Description

List of proposed changes:

- Added status bar to the universal card component
- Updates color and text based on the inputted status prop


components/atoms/StatusBadge.js
```diff
@@ -6,9 +6,18 @@ import propTypes from 'prop-types'
...
+        className={`font-medium font-display text-black text-lg px-5 py-1 sm:mb-5 w-full sm:w-1/3 sm:mr-14 rounded-t-lg sm:rounded-t-none sm:rounded-b-lg 
+        ${
+          props.status === 'In payment'
+            ? 'bg-status-inPayment'
+            : props.status === 'Benefit update'
+            ? 'bg-status-benefitUpdate'
+            : props.status === 'Application received'
+            ? 'bg-status-applicationReceived'
+            : 'bg-status-inactive'
+        }`}
```
s-laugh, line 20: This won't work if you translate to french. 
One option to fix it would be to pass the code to the component. Then this check can be made using the code (importing the code const). And you would then load the translation from the translation file based on the code.

s-laugh (changes requested):

components/atoms/StatusBadge.js
```diff
@@ -25,7 +25,7 @@ export default function StatusBadge(props) {
             : 'bg-status-inactive'
         }`}
       >
-        {t[props.status]}
+        {t[props.status] ? t[props.status] : 'No status found'}
```
s-laugh, line 28: Just a slightly cleaner way to get the same result (only works because it's not calling a sub-property of the element you are checking)
```suggestion
        {t[props.status] ?? 'No status found'}
```

PR #317, opened 2022-05-04
add the expand hide feature to new card component

s-laugh:
### Description

List of proposed changes:

- Copies over the ViewMoreLessButton component and the functions to expand hide the task list section from the BenefitCard.
- Uses the task list mapping to determine what the text of the button should be
  - Note: going forward (from design review) this is only going to have one text and not change base on state, so I didn't add that functionality here
  - Additional Note: I used the mapping like this, as that text will change based on program and status (but also be repeated). So re-used the same map find function to not duplicate text in the translation.

### DIFF
--- a/components/molecules/UniversalBenefitCard.js
+++ b/components/molecules/UniversalBenefitCard.js
@@ -5,12 +5,30 @@ import { ProgramCodes, StatusCodes } from '../../objects/UniversalBenefit'
 import BenefitCardHeaderSummary from './BenefitCardHeaderSummary'
 import BenefitTasks from './BenefitTasks'
 import HorizontalRule from '../atoms/HorizontalRule'
+import { useState, useRef } from 'react'
+import ViewMoreLessButton from '../atoms/ViewMoreLessButton'
 
 export default function UniversalBenefitCard(props) {
   const t = props.locale === 'en' ? en : fr
+  const [isOpen, setIsOpen] = useState(false)
+  const topOfCardRef = useRef(null)
+  const topOfTaskRef = useRef(null)
+
+  const scrollTo = () => {
+    if (!topOfTaskRef.current || !topOfCardRef.current) return
+    isOpen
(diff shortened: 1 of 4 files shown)


components/molecules/UniversalBenefitCard.js
```diff
@@ -5,12 +5,30 @@ import { ProgramCodes, StatusCodes } from '../../objects/UniversalBenefit'
...
 import BenefitTasks from './BenefitTasks'
 import HorizontalRule from '../atoms/HorizontalRule'
+import { useState, useRef } from 'react'
+import ViewMoreLessButton from '../atoms/ViewMoreLessButton'
 
 export default function UniversalBenefitCard(props) {
   const t = props.locale === 'en' ? en : fr
+  const [isOpen, setIsOpen] = useState(false)
+  const topOfCardRef = useRef(null)
+  const topOfTaskRef = useRef(null)
```
DaceyTom2, line 15: Avoid using ref if possible

components/molecules/UniversalBenefitCard.js
```diff
@@ -24,20 +42,34 @@ export default function UniversalBenefitCard(props) {
...
+                    />
+                    <HorizontalRule width="w-auto sm:w-full" />
+                  </div>
+                )
+              })}
+          <ViewMoreLessButton
+            id={props.benefit.programCode + '-card-button'}
+            onClick={() => {
+              setIsOpen(!isOpen)
+              scrollTo()
```
DaceyTom2, line 66: instead of a scrollTo function try looking into calling scrollIntoView directly
parent.scrollIntoView()

CurtisUnderwood (approved):
Nice work, lgtm. We should note for future reference the BenefitTasksConstants will need to take a dynamic locale value since it's still taking const t = en.

s-laugh:
Merging as this seems to be a blocker. @DaceyTom2 please feel free to double check and I'll be happy to fix anything in another PR.

PR #319, opened 2022-05-05
Server side auth

DaceyTom2:
### Description
Anthan added custom provider for authentication.
This is a PoC that will be replaced with another provider when ready

List of proposed changes:
- Can authenticate with Keycloak temporarily
- Can logout through the Dashboard page

/api/auth/signin to authenticate
Click "sign out" on the /dashboard page to sign out


nibivi77 (approved):
Nice work! :)

PR #321, opened 2022-05-06
Replacing Account Details with Task in list

s-laugh:
### Description

- Removes the account details sections and related components/text
- Adds a task to most cards to link to profile page

#### Additional note
Tasks list groups still need to be updated to match designs


nibivi77 (approved):
:+1: Looks good

PR #322, opened 2022-05-06
DCWC-1708 Formatted summary data

CurtisUnderwood:
### Description

List of proposed changes:

- Formatted benefit summary data on object creation

### Additional Notes
- Avoids adding extra logic in the component to keep the component as dynamic as possible
- Removed spacing on $ format in translation files since it's not available in the designs


objects/UniversalBenefit.js
```diff
@@ -144,9 +146,9 @@ export function CreateBenefitObjWithCPPData(cppBenefitData) {
     [
       CreateBenefitSummary(
         SummaryTypes.PaymentAmount,
-        cppBenefitData.netAmount
+        '$' + cppBenefitData.netAmount
```
s-laugh, line 149: The $ sign would need to go after the number in french. 
We probably want to handle this on the component. You can see how it's done here as an example https://github.com/DTS-STN/sc-digital-centre/blob/cc0046397c76bb8b24b588feba157a7fc4e3ccec/components/atoms/BenefitCardHeaderActive.js#L90

objects/UniversalBenefit.js
```diff
@@ -144,9 +146,9 @@ export function CreateBenefitObjWithCPPData(cppBenefitData) {
...
       ),
-      CreateBenefitSummary(SummaryTypes.NextPayment, nextPayment),
+      CreateBenefitSummary(SummaryTypes.NextPayment, formatDate(nextPayment)),
```
s-laugh, line 151: While this might seem simpler, I believe the formatting/design should be managed in the component. This requires us to format the date each time we load it into the object, which could happen many times. Alternatively formatting it when it's loaded in the component, means we only need to call the "formatDate" once.
It's probably going to require a Switch/Case on the SummaryType to get the right format for the type of data.

s-laugh (changes requested):
As well as the comments, don't forget the text size, weight, and colour change based on the summaryType in some cases. You should be able to use a single switch/case logic to apply a data format and a css class.

objects/UniversalBenefit.js
CurtisUnderwood: Adjusted the font-weight by summaryType, don't think there are any more changes to text size/colour based on the summaryType in the current iteration of the master components but if there's anything missing let me know.

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -1,15 +1,42 @@
...
+    }
+  }
+
   return (
     <div>
       <p>{t[props.summary.type].title}</p>
-      <p className="font-bold">{props.summary.value}</p>
+      <p
+        className={
+          props.summary.type === SummaryTypes.PaymentAmount
```
s-laugh, line 33: Would it be possible to re-use the same case/switch to define both the value and the style class at the same time?

s-laugh (commented):
Looking good; double check font weight, colour, and size for all the elements (title, data, link) that they match the EI & CPP Active cards (as that's what these are mimicking)

PR #325, opened 2022-05-09
added styles to the card header part

shewood:
### Description

List of proposed changes:
Updated the styling on the benefit card header.
-
-


components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -7,18 +7,20 @@ export default function BenefitCardHeaderSummary(props) {
...
-      {!typesWithLinks.find((t) => t === props.summary.type) ? null : (
-        <p>
-          <Link
-            text={t[props.summary.type].linkText}
-            href={t[props.summary.type].link}
-          />
-        </p>
-      )}
+    <div className="w-full ">
+      <div className="w-full  ">
```
s-laugh, line 11: There's a duplicated div here.

s-laugh (commented):
Can you also merge dev into this, will fix the task lists, so we can verify they work nicely with this design update. Other than that it's great!

components/molecules/BenefitCardHeaderSummary.js
shewood: Yes, I saw that and i will update. Thanks

components/molecules/BenefitCardHeaderSummary.js
```diff
@@ -7,12 +7,12 @@ export default function BenefitCardHeaderSummary(props) {
   const typesWithLinks = [SummaryTypes.PaymentAmount, SummaryTypes.LatestStatus]
 
   return (
-    <div>
-      <p>{t[props.summary.type].title}</p>
-      <p className="font-bold">{props.summary.value}</p>
-      <p className="font-bold">{props.summary.status ?? null}</p>
+    <div className="w-full">
+      <p className="">{t[props.summary.type].title}</p>
```
DaceyTom2, line 11: className can be removed

s-laugh:
Don't forget about the expanded section of the card - still needs some work.
Also be careful about editing those shared components, the text on some of the view more button looks off.
image

DaceyTom2 (approved):
Small change but looks great

shewood:
I will take a look at the expanded section. I wasn't looking at that, just the header for now.

s-laugh:
> I will take a look at the expanded section. I wasn't looking at that, just the header for now.

It's more the expanded section and view more were looking good in dev, so something is conflicting in the styles here.

components/molecules/BenefitCardHeaderSummary.js
shewood: fixed

shewood:
I updated the placement of the taskList. Now it shows under the button as expected.

s-laugh (approved):
Looks Good!

PR #330, opened 2022-05-11
Configuring auth for server

DaceyTom2:
### Description
Enabling auth to work on the server

List of proposed changes:

- Added env variables to docker
- Added env variables to teamcity build

Reach dashboard page and authenticate
If unauthenticated you should be kicked to login screen
Once authenticated the dashboard header sign out button should log you out


nibivi77 (approved):
Nice! This looks like it going to be cool to use.

PR #333, opened 2022-05-11
changing the API to actually be an API

s-laugh:
### Description
Fetches the benefits data from introp stub in a dynamic way (via client)
Removes advertising cards (rather than not loading them)


pages/dashboard.js
```diff
@@ -63,10 +63,102 @@ import UniversalBenefitCard from '../components/molecules/UniversalBenefitCard'
...
+    [usersBenefits, FilterAdvertisingCards]
+  )
+
+  useEffect(() => {
+    fetchProgramData('cpp', setCppLoading, setCppError)
+  }, [fetchProgramData])
+
+  useEffect(() => {
+    fetchProgramData('ei', setEiLoading, setEiError)
+  }, [fetchProgramData])
```
DaceyTom2, line 160: The useEffect can be combined into one with empty brackets at the end ie [fetchProgramData] -> []
The app should only make this request once and I'm concerned about the behaviour of setting it to a call back.

Ideally useEffects should be triggered once on page load or multiple times triggered by updated state

DaceyTom2 (changes requested):

pages/dashboard.js
s-laugh: It could be change it to call both API's in one `useEffect`, but thought you had suggested it be broken into two?

The brackets with `[fetchProgramData]` is required to identify the dependency of that function. And that function is using a `useCallback' because it's also required for compile as the function is referenced in the useEffect. Changing the useCallback function to be a regular function does cause the recursive infinite loop of API requests (or adding duplicate from the result - I didn't debug as to which it was)

DaceyTom2: Take some time to read over the ReactJS documentation. Specifically the React hook documentation on useState and useEffect.

There are also examples of querying using useState and useEffect in scrum poker

s-laugh: So reading the docs, and looking at this change, I'm still not fully clear on what your suggesting. That being said, I've pushed a change that moves both API requests into one useEffect and the API fetch into an async function within it. (The issue I think you are trying to point out being that they are both updating a dependency of `userBenefits` of each other, so they could keep re-triggering the other once one completes).

I did leave the FilterAdvertisingCards as a useCallback, this seems to be agreeable to the docs; and everything else I tried broke the app.

DaceyTom2 (approved):
Approving and merging this PR to push it through. @s-laugh is away this week and the PR is in good shape

pages/dashboard.js
DaceyTom2: Merging as code is needed for other sprint jira tickets

PR #334, opened 2022-05-11
fixing some errors caused by wrong propTypes

s-laugh:
fixing the propTypes definitions and some other tweaks


s-laugh:
> proptype errors, Is this just to fix one set?

I do see on other, but it's in the Design System, is that what you're getting too?

CurtisUnderwood (approved):
Nice work!

shewood (approved):
Looks Good!!

PR #337, opened 2022-05-12
adding CPP mocks and adjusting summaries for them

s-laugh:
### Description

List of proposed changes:

-
-


CurtisUnderwood (changes requested):
Should we be mocking all data in the mock? For example, this doesn't include the In Payment card in CPPMockData,js which is getting its data from somewhere else? I feel it might make things confusing

s-laugh:
> Should we be mocking all data in the mock? For example, this doesn't include the In Payment card in CPPMockData,js which is getting its data from somewhere else? I feel it might make things confusing

Yup, probably a good idea.

Issue #320, opened 2022-05-05
Maybe the LayoutContainer component and this div should be moved to the Layout component, so that it's on every page (fyi the `col-span-12` isn't needed anymore)?


s-laugh:
Maybe the LayoutContainer component and this div should be moved to the Layout component, so that it's on every page (fyi the `col-span-12` isn't needed anymore)?

_Originally posted by @s-laugh in https://github.com/DTS-STN/sc-digital-centre/pull/318#discussion_r865844540_

PR #338, opened 2022-05-12
Layout container

s-laugh:
### Description

fixing #320 

List of proposed changes:

- Moved LayoutContainer implementation to the Layout component for each page
- Removed the use of it from individual pages along with redundant divs

### Additional Notes

None of the content of the pages was edited at all, just the line indenting.


CurtisUnderwood (changes requested):
image

Some issues with the splash page. Looks like it's still taking the footer somehow and the background image isn't stretching the full page

s-laugh:
> Some issues with the splash page. Looks like it's still taking the footer somehow and the background image isn't stretching the full page

Funny enough, the footer has been there for a while... both issues fixed now.

components/organisms/Layout.js
```diff
@@ -78,7 +79,11 @@ export default function Layout(props) {
       )}
 
       <main>
-        <div>{props.children}</div>
+        {props.fullscreen ? (
+          props.children
+        ) : (
+          <LayoutContainer>{props.children}</LayoutContainer>
+        )}
```
DaceyTom2, line 86: I think my review got deleted but layout and layoutContainer should be applied to _app so they don't have to be added to pages

DaceyTom2 (changes requested):

components/organisms/Layout.js
s-laugh: Wasn't quite in the scope of what I was aiming to do, but I've added that change as well.

DaceyTom2 (approved):
Looks great!

CurtisUnderwood (approved):
Nice work!

PR #339, opened 2022-05-12
fix for the overflowing text

shewood:
### Description
The text in the "You might be eligible" card didn't wrap and was extending over the icons and causing mobile to have a horizontal scroll bar. 
List of proposed changes:
The link text now wraps and stays in the designated area. 
the LInk text is on the line below the other text.
made the paragraph text a p tag from an a tag as it didn't appear to be a link.
added an h2 for the card


DaceyTom2 (approved):
Looks way better!

PR #347, opened 2022-05-18
Dcwc 1774 remove static benefit cards

DaceyTom2:
### Description
All benefit cards are dynamic now. This should make it easier to mock cases out.

List of proposed changes:
- Benefits are mapped
- List of benefits come from content

Ensure page still works the same


shewood (approved):
Looks good.

PR #351, opened 2022-05-19
small background color change

nibivi77:
Small change for styling. (I forgot to extend the background color all the way across.)


shewood (approved):
Looks good.

PR #356, opened 2022-05-26
updating the heading hierarchy

s-laugh:
### Description

List of proposed changes:

- Updating the headings so they have a screen reader flow intend by client
- Updated a few elements to lists so screen reader will more easily iterate though them
- Added some translated screen reader only text

### Additional Notes

Edits to the BenefitCard component were not supper well thought out out as it's expected to be removed, so were just to make the tests pass.


DaceyTom2 (approved):
Looks good

PR #358, opened 2022-05-26
DCWC-1836 Scenario mock id example

DaceyTom2:
### Description
Scenario mock id example with CPPD

List of proposed changes:
- Set scenario id
- Mock file structure added with minimal examples

### Additional Notes
I'm unsure what the CPPD program and benefit codes are. I left them as CPP for now
  "programCode": 32294,
  "benefitCode": 30320,


s-laugh (approved):
Very nice!

PR #360, opened 2022-05-26
update cpp mocks

s-laugh:
### Description

List of proposed changes:

- Mocks the Introp json result instead of the clients result
- Updates one of the functions names as agreed.


CurtisUnderwood (changes requested):
The userid cookie has a value of null so the if statement in cpp.js isn't firing and is not loading the default or any of the user-specific mocks?

CurtisUnderwood (approved):
Previous comment was referring to something that will be resolved in another PR, LGTM.

PR #362, opened 2022-05-31
Update english content

nibivi77:
### Description (Important notes!)
- I only modified our Universal Benefit card component as we will want to move forward with this component.

### DIFF
--- a/components/molecules/UniversalBenefitCard.js
+++ b/components/molecules/UniversalBenefitCard.js
@@ -11,6 +11,9 @@ import ViewMoreLessButton from '../atoms/ViewMoreLessButton'
 export default function UniversalBenefitCard(props) {
   const t = props.locale === 'en' ? en : fr
   const [isOpen, setIsOpen] = useState(false)
+  const [btnCaption, setBtnCaption] = useState(
+    props.benefit.taskHeadingKey ? t[props.benefit.taskHeadingKey] : t.viewMore
+  )
   const benefitUniqueId =
     props.benefit.programCode +
     '-' +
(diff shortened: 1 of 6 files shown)


locales/en.js
```diff
@@ -352,8 +351,10 @@ export default {
   commonActions: 'Common actions',
 
   //benefit task headings
-  commonActions: 'Explore common actions',
-  paymentsTaxesAccount: 'Payments, taxes and account information',
+  viewMore: 'Explore common actions',
```
s-laugh, line 354: I think this could use a better prop name.

s-laugh (commented):
Looks good, I think we just need to confirm what we can do with the viewMoreLess button to ensure we meet a11y

components/molecules/UniversalBenefitCard.js
```diff
@@ -96,11 +99,20 @@ export default function UniversalBenefitCard(props) {
               behavior: 'smooth',
               block: 'start',
             })
+            setBtnCaption(
+              isOpen
```
s-laugh, line 103: _My comment didn't post_
As far as I know, to meet a11y we can't change a heading based on a state. (the caption is used as a h5 I believe)

nibivi77: Ok, how else should we deal with changing the text depending on the benefit and status?

Do we know exactly what the key/value pairs will be from the benefits objects returned from our api?

locales/en.js
nibivi77: changed

components/molecules/UniversalBenefitCard.js
s-laugh: Sorry what I meant by "state" was open or closed, as in the user did something and the text in the header changed, that's not allowed. So when it's open, we can't change the text to "view less" (that's also insufficient at heading text, as it's not descriptive of the content within the section)

So all I'm suggesting is we remove the condition on `isOpen` and only use the text  reflective of program+status for both open and closed states (this is a pretty common thing on web pages).

I've mentioned the issue with Breanna last week, just have to confirm with design/content.

nibivi77: Right, so that should be brought up to the design team. Let's discuss this with them on stand up.

s-laugh (approved):
Looks good!

PR #363, opened 2022-05-31
removing BenefitCard component and associated parts

s-laugh:
### Description

List of proposed changes:

- Removed a number of no longer needed components
- Removed and refined some logic that was used by the removed components
- Cleaned up some other code where bugs popped up from the removal.


shewood (approved):
I tried it out and it looks good.

PR #365, opened 2022-06-01
DCWC-1839 - Create Mock Scenario 3

CurtisUnderwood:
### Description

List of proposed changes:

- Added scenario 3
- Bug fix on forEach being called on undefined if no mock data is passed

### Additional Notes
- Are tests needed for the scenarios?


shewood (approved):
Looks good. The error is fixed.

PR #366, opened 2022-06-01
Add All Status Mocks for CPPD

s-laugh:
### Description

List of proposed changes:

- configure CPPD API to accept/return an array
- combine CPP and CPPD mapping, reducing code duplication
- added default mocks for all status' of CPPD


CurtisUnderwood (approved):
Nice work

PR #368, opened 2022-06-02
DCWC-1838 Create Mock Scenario 2

CurtisUnderwood:
### Description

List of proposed changes:

- Added mock test scenario 2
- Currently API response for Latest status update data is unknown so it is passed in plain text until we have more information on this


s-laugh (approved):
Looks good!

PR #369, opened 2022-06-02
temporary mock instance of gis

nibivi77:
### Description

This is only mocking one status of GIS.
Adding more might not be a good idea since we don't know what the object will look like when receiving it.

Mapping it will come once we have an api call for it.

Correct me if I am wrong, please.


CurtisUnderwood (changes requested):
Looks good, GIS.js should get the userData cookie and return benefits from a mock data like in cpp.js as well.

`let userData
      const userid = getCookie('userid', { req, res })
      if (userid) {
        //Mock userid response
        userData = mockData[userid].CPP
        }`

PR #370, opened 2022-06-02
Common functions

s-laugh:
[DCWC-1886](https://jira-dev.bdm-dev.dts-stn.com/browse/DCWC-1886)

### Description

List of proposed changes:

- Reduces a lot of duplicated code in program data fetch
- Properly handles errors (todo: translate and design errors)

### Notes

Programs that don't have an Interop stub, should show as a 501 - Not Implemented error when on the dashboard without a userid


pages/api/programData/_middleware.js
```diff
@@ -15,3 +17,46 @@ export async function middleware(req, ev) {
     }
   }
 }
+
+export async function GetProgramData(req, res, getPath, getMockObj, mapCard) {
```
DaceyTom2, line 21: This logic belongs in /lib

s-laugh: thanks for the tip - I was also thinking similarly

TishamIslam (approved):
Looks great! Glad to see all that duplicate code getting put in a single location.

PR #371, opened 2022-06-02
Mock all versions of the OAS cards

TishamIslam:
### Description
Adds all the versions of the OAS cards when a userid of `"default"` is entered

List of proposed changes:

- Changed OAS api to return an array and map it on the dashboard
- Added mapping codes for OAS
- Added a `ProgramTypeCode` for OAS `"OASBeneficial"`
- Added `MapOASCard()` to `BenefitsMapping.js`
- Added Mock data for oas
- `userid` of `"default"` won't throw an error anymore for `cppd.js` api


pages/api/programData/cppd.js
```diff
@@ -7,7 +7,8 @@ export default async function handler(req, res) {
     //get data
     let userData
     const userid = getCookie('userid', { req, res })
-    if (userid) {
+    if (userid != 'default') {
```
s-laugh, line 10: this should get reverted before merging

pages/api/programData/oas.js
```diff
@@ -1,19 +1,42 @@
...
+        userData.forEach((result) => {
+          benefits.push(MapOASCard(result))
+        })
+      } else {
+        //handle the other situations here
+      }
+      res.status(200).json(benefits)
+    } else {
+      //Otherwise send this mock
+      res.status(200).json([
```
s-laugh, line 25: I don't see a need for us to keep this mock anymore, am I missing something?

s-laugh (changes requested):
Great work! Just some cleanup needed.

pages/api/programData/oas.js
TishamIslam: Ah that's true

s-laugh (approved):

PR #374, opened 2022-06-06
Dcwc 1768 gis temp mock

nibivi77:
### Description

Adding all mocked states for GIS cards.

It should be exactly like OAS.


locales/en.js
```diff
@@ -1,5 +1,12 @@
 /* eslint import/no-anonymous-default-export: [2, {"allowObject": true}] */
 export default {
+  //
+  // Benefits
+  CPP: 'Canada Pension Plan',
+  OAS: 'Old Age Security',
+  GIS: 'Guranteed Income Supplement',
+  EI: 'Employment Insurance',
+  CPPD: 'Canada Pension Plan Disability',
```
s-laugh, line 9: These are duplicated with a lowercase instance, GIS included https://github.com/DTS-STN/sc-digital-centre/blob/aecb65a491f239855d25156c3d6eecefa6f99967/locales/en.js#L289
I believe we are now only using the lowercase instances.

s-laugh (approved):
Awesome!

PR #375, opened 2022-06-06
Fixed errors and cppd task list bug

CurtisUnderwood:
### Description

List of proposed changes:

- Fixed summaries as a required prop when not all cards take summaries
- Fixed issue with CPPD task list not expanding


s-laugh (approved):
Good finds!

PR #380, opened 2022-06-07
Cards will now have localized date and currency strings

TishamIslam:
### Description
Adds localisation for dates and currencies on the cards. Also added a `formatMoney()` method for future convenience

List of proposed changes:

- Localisation of dates and currencies
- Removed currency strings from locale files as they're no longer needed


s-laugh (approved):
Looks Great!

s-laugh:
Merging to include in sprint.

PR #381, opened 2022-06-08
Task and Translation Alignment

CurtisUnderwood:
### Description

List of proposed changes:

- Aligned tasks and task headers with master components
- Added some missing translations (WIP)

### DIFF
--- a/components/atoms/ViewMoreLessButton.js
+++ b/components/atoms/ViewMoreLessButton.js
@@ -4,7 +4,7 @@ import ActionButton from './ActionButton'
 export default function ViewMoreLessButton(props) {
   return (
     <ActionButton
-      className="text-xl leading-8 py-5 pl-2 md:pl-6 underline text-link-blue-default hover:text-link-blue-hover"
+      className="text-xl leading-8 py-5 pl-2 md:pl-6 text-link-blue-default hover:text-link-blue-hover"
       onClick={props.onClick}
       type="button"
       id={props.id}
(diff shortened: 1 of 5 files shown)


components/atoms/ViewMoreLessButton.js
```diff
@@ -4,7 +4,7 @@ import ActionButton from './ActionButton'
 export default function ViewMoreLessButton(props) {
   return (
     <ActionButton
-      className="text-xl leading-8 py-5 pl-2 md:pl-6 underline text-link-blue-default hover:text-link-blue-hover"
+      className="text-xl leading-8 py-5 pl-2 md:pl-6 text-link-blue-default hover:text-link-blue-hover"
```
s-laugh, line 7: I believe this would make that button/link inaccessible... (it's a bit subjective, so maybe we should check with an a11y expert)
https://www.w3.org/WAI/WCAG21/Techniques/failures/F73
> Check that each link in the page that is identifiable by color (hue) is visually identifiable via some other means (e.g., underlined, bolded, italicized, sufficient difference in lightness, etc).

shewood (commented):
Was there a design to remove the underline from the viewmorelessbutton?

CurtisUnderwood:
> Was there a design to remove the underline from the viewmorelessbutton?

Seems like they were removed in all the latest master components.
image

shewood (approved):
Looks good.

PR #383, opened 2022-06-09
Update to Key/Value pair look-up

s-laugh:
### Description

List of proposed changes:

- updating look-ups for prop reference
- removing unused task lists (bonus)

Validate that all the cards still render without error

### Additional Notes

This kind of lookup can be more error prone, but more common in a key/value pair model which is way faster (even if in this instance the speed is negligible). The look-ups are being done within a try catch of the API fetch so will be handled there if Interop passes us unexpected data. This is in line with a "fail fast" programming style (we want' the code to break if something unexpected happens, so we take notice sooner and can fix it faster)


DaceyTom2 (approved):
Huge improvement!

PR #390, opened 2022-06-14
Add API Tests

s-laugh:
### Description

List of proposed changes:

- adding cpp api test
-

### Additional Notes
Adds new dev dependencies to mock stuff


s-laugh:
> I think this looks pretty good, but it looks like there's a bit of duplicated stuff here. Is there any way to parameterize these tests so that we could, for example, run the CPPD, OAS, GIS and SEB tests as one test while passing in the program codes and handlers as parameters?

In general, we don't want to edit code to run a test - unless that code is untestable because it's not allowing injection so we can include mocks. The other thing to consider here is leveraging the tests to validate that changes to those common functions don't unexpectedly change other programs; keeping the tests individual to the program adds that layer of validation.

s-laugh:
To add clarity to my statement, what's being tested here is the `await handler(req, res)`, so there is a test for each one.
In the case of EI and CPP we have a env URL we are passing, so we need to validate that it's being used, but to be clear what it's testing is this line https://github.com/DTS-STN/sc-digital-centre/blob/28b2428028f90bf51fecfd221d99480b1ddcb04b/pages/api/programData/ei.js#L8

We are NOT intentionally testing the https://github.com/DTS-STN/sc-digital-centre/blob/dev/lib/api/programData.js with these.

nibivi77 (approved):
Well done Shaun!

PR #391, opened 2022-06-14
Dcwc 1875

shewood:
### Description

List of proposed changes:
Removed some unused colours from the tailwind config file to help reduce the build file size. 
Renamed a few colours and added in the specific colours from figma. 
renamed link-blue to just blue- and added the states for hover, default and pressed as this is also applicable for buttons.
-

Check that nothing broke.


s-laugh:
Looks like the English and French links on the splash page are different... I don't see the code that change it though.

shewood:
> Looks like the English and French buttons on the splash page are different... I don't see the code that change it though.

Good catch. I added the primary blue colour that's used a lot on canada.ca page.

s-laugh (approved):
Awesome

PR #394, opened 2022-06-14
1926 translation header summary

DaceyTom2:
### Description
Removed translations from header summary component

### Additional Notes
Logic temporarily moved to UniversalBenefitCard


DaceyTom2:
> Looks good. FYI take a look at the UniversalBenefitCard test when your done with these, I noticed a propType error.

Will do, Trying to do this one component at a time

PR #398, opened 2022-06-15
DCWC-1877 Pass state id in language toggle link

CurtisUnderwood:
### Description

List of proposed changes:

- Added userid to langToggleLink

### Additional Notes
- There is an existing issue where users can leave to the profile page and return to the dashboard to have their cookie wiped, may need to be resolved if scenarios are being demo'd

### DIFF
--- a/pages/dashboard.js
+++ b/pages/dashboard.js
@@ -204,7 +204,14 @@ export async function getServerSideProps({ req, res, locale, query }) {
     description: 'en + fr description',
   }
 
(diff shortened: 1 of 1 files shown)


s-laugh (approved):
👍 

If you're feeling up for it, think of how this could be done in a way that doesn't repeat conditions (`locale === 'en' ?...`) and doesn't repeat the hardcoded text (`'/fr/dashboard'`) or use any hardcoded text.

s-laugh (approved):
Well done!

PR #399, opened 2022-06-16
Dcwc 1930 translate benefit tasks

DaceyTom2:
### Description
Removed translations from benefit tasks

List of proposed changes:
- Translations added to content file as will be coming from AEM

Ensure app runs similarly to before


s-laugh (approved):
:+1:

PR #400, opened 2022-06-17
Dcwc 1931

shewood:
### Description
Commented out the code for the no benefit cards - 
Added tests for error pages
Updated tests on dashboard page for A11y - found a command that gives more info about the error is running in the cypress runner. 
started tests for other pages - still in progress -


cypress/integration/dashboard.spec.js
```diff
@@ -1,18 +1,37 @@
...
+    it('dashboard page is accessible', () => {
+      cy.checkPageA11y('/dashboard')
+    })
+  });
+
+describe('The dashboard page loads', () => {
     beforeEach(() => {
       cy.visit('/dashboard')
-    //   cy.injectAxe();
+       cy.injectAxe();
```
nibivi77, line 12: `checkPageA11y` is already injecting Axe I think, right?

or maybe it's unique to each `describe` block. If so, do we need both `describe` blocks? can we merge them into one?

shewood: Normally it would be only one but I put them separate as I am ignoring the error to catch the 400 and I didn't want that always.

cypress/support/commands.js
```diff
@@ -23,3 +23,43 @@
...
+          name: '🔧',
+          consoleProps: () => violation,
+          $el: Cypress.$(target.join(',')),
+          message: target
+        })
+      })
+    });
+  }
+  
+  Cypress.Commands.add("checkPageA11y", (path) => {
```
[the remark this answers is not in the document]
shewood: I wasn't sure as one ignores the error and I didn't think we wanted that always. Any suggestions?

cypress/integration/dashboard.spec.js
nibivi77: Ok fair enough! If the team has thoughts on this we can come back to it!

cypress/support/commands.js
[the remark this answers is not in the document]
nibivi77: Ok fair enough! If the team has thoughts on this we can come back to it!

nibivi77 (approved):
If the team wants to take a look at the resolved conversations they can! Good work!

PR #403, opened 2022-06-20
Consolidate authentication for reusability

s-laugh:
### Description

List of proposed changes:

- Bringing the authentication checks of session into one spot to enable better reusability
- Set profile page to authenticate with new functions, converting to server page.


DaceyTom2 (commented):
I'd recommend avoiding creating wrappers for libraries

s-laugh:
> I'd recommend avoiding creating wrappers for libraries

Generally I agree, I feel like this makes for a good use case for an exception; as this is mocking a session object when we want to "bypass" the authentication.

Would like to hear more of your thoughts around this

DaceyTom2:
bypassing authentication is just temporary until we get auth working with ECAS. The bypass should be removed after

s-laugh:
Yup, not sure there is even much need for the bypass now... we have the dev keycloak for that.

PR #404, opened 2022-06-20
check for undefined summaries

s-laugh:
### Description

List of proposed changes:

- make a check for undefined summaries before executing a function on them

 - https://sc-digital-centre-dev.bdm-dev.dts-stn.com/dashboard?userid=default (loads until it crasshes)


DaceyTom2 (approved):
Good catch

PR #409, opened 2022-06-22
DCWC-1947 added logic for no benefit cards

CurtisUnderwood:
### Description

List of proposed changes:

- No Benefit Cards now only display when userid=default

### Additional Notes
- This could be done by passing userid as a prop as well but was done with the cookie in order to not duplicate data available to the client


s-laugh (commented):
Great! would you be able to add a test for it?

__tests__/pages/dashboard.test.js
```diff
@@ -14,7 +16,24 @@ expect.extend(toHaveNoViolations)
 jest.mock('next-auth/react')
 enableFetchMocks()
 
+jest.mock('cookies-next', () => ({
+  getCookie: () => 'default',
```
s-laugh, line 20: This doesn't look to be doing quite what we expect; in the coverage report it's showing that this line (https://dts-stn.github.io/sc-digital-centre/refs/pull/409/merge/coverage/lcov-report/pages/dashboard.js.html#L275) isn't getting hit; which it should if getCookie returns `default`. Line 55 has an example of how a return value of getSession is being mocked.

__tests__/pages/dashboard.test.js
```diff
@@ -77,6 +96,10 @@ describe('Dashboard', () => {
     expect(cppdResult).toBeInTheDocument()
   })
 
+  it('renders NoBenefitCards', () => {
+    expect(NoBenefitCards).toBeTruthy()
```
s-laugh, line 100: Unfortunately all this is doing is testing the code from the const, which is in the test file. If the dashboard changes this will remain completely un-effected. The const NoBenefitCards really isn't needed.

s-laugh (changes requested):
Let me know if you want to chat about this, I have some time to work though it with you.

__tests__/pages/dashboard.test.js
s-laugh: nvm, this is working; coverage wan't updated when I looked at it.

PR #410, opened 2022-06-22
updated the colours as per DS

shewood:
### Description

List of proposed changes:
It was desired to have the colours come from the DS has so some had to be adjusted slightly. 
The lighter palette was chosen as per figma


TishamIslam (approved):
Looks good!

PR #411, opened 2022-06-23
1949 cardheader update

s-laugh:
### Description

List of proposed changes:

- Adds a required "callout" section to the card header
- Updates the design of the card header
- Updates the summaries of the card header to mach design (which elements are shown)

Validate that everything is working as expected, and no missing parts.

### Additional Notes

This sets what should be the correct padding/margin around the "apply" buttons, but the icon until updated makes it look too big.


constants/EIMappingCodes.js
```diff
@@ -2,7 +2,7 @@ import { TypeCodes } from './ProgramTypeCodes'
 import { StatusCodes } from './StatusCodes'
 
 export const EITypes = {
-  1: TypeCodes.EIUnknown,
+  1: TypeCodes.EISickness,
```
TishamIslam, line 5: Why do we have 2 TypeCodes with the same value?

shewood:
You might want to consider sm:gap-2 or more so that the text for requested benefit doesn't touch the status update date with those massive fonts.

s-laugh:
> You might want to consider sm:gap-2 or more so that the text for requested benefit doesn't touch the status update date with those massive fonts.

Great suggestion! I used a lg:gap-7 to get a 28px as the design grid shows gaps of ~30px; but also a sm:gap-4 because imo it was too much on smaller screens.

shewood (approved):
The gap change looks great.

PR #412, opened 2022-06-23
Dcwc 1960 estimated decision date

TishamIslam:
### Description
Changes up the estimated decision dates based on this [table, under the "Estimate Decision Date Logic" tab](https://014gc.sharepoint.com/:x:/r/sites/DECD/_layouts/15/Doc.aspx?sourcedoc=%7BABFCF0A7-6257-4D36-AF51-BC2C2912EBC0%7D&file=Backlog%20Overview.xlsx&action=default&mobileredirect=true)
This also removes the "Estimated decision date" from cards with the status "Decision letter sent" as the estimate is irrelevant once the decision has been made.  

List of proposed changes:

- Estimated decision date removed from cards with "Decision letter sent"
- Estimated decision date removed from cards if the date is in the past.
- Estimated decision date varies between benefits now
- `formatDate()` util method now handles millisecond inputs (no more invalid date on the UI)
- Used `formatDate()` on date data in `mapBenefits.js` to prevent dates from being off by one day

Make sure tests run properly, look for any other "Invalid date" messages on the UI or anything similar.

### Additional Notes
Not sure about using `formatDate()` in `mapBenefits.js`, might be better to just append `T12:00:00.000Z` to each date string?


lib/api/mapBenefits.js
```diff
@@ -77,11 +80,15 @@ export function MapEICard(eiBenefitData) {
...
           SummaryTypes.EstimatedDecisionDate,
-          estimatedDateOfDecision.setMonth(
-            estimatedDateOfDecision.getMonth() + 1
+          estimatedDateOfDecision.setDate(
+            estimatedDateOfDecision.getDate() + 28
           )
         ),
       ]
+      //remove estimated decision date if current day is past that point
+      if (Date.now() > estimatedDateOfDecision.getTime()) {
```
s-laugh, line 89: This check would be better done, before adding the date. Check for other spots where this could be improved as well.

lib/api/mapBenefits.test.js
```diff
@@ -14,8 +14,18 @@ import SEBMockData from '../../mockdata/userDefault/seb.json'
 import { StatusCodes } from '../../constants/StatusCodes'
 import { SummaryTypes } from '../../constants/SummaryTypes'
 
-describe('maps EI', () => {
-  it('Application Received status', () => {
+const options = { year: 'numeric', month: 'numeric', day: 'numeric' }
+const timezone = 'T12:00:00.000Z'
+
+describe('Maps EI', () => {
+  it('Application Received status - Without estimated decision', () => {
```
s-laugh, line 21: This doesn't actually check that the estimated decision is NOT returned. Again, check for other places where this applies.

lib/api/mapBenefits.test.js
```diff
@@ -14,8 +14,18 @@ import SEBMockData from '../../mockdata/userDefault/seb.json'
...
+    const mappedData = MapEICard(EIMockData[0])
+    expect(mappedData.statusCode).toBe(StatusCodes.applicationReceived)
+    expect(mappedData.summaries).toHaveLength(2)
+  })
+
+  it('Application Received status - With estimated decision', () => {
```
s-laugh, line 27: Also here, this doesn't actually check to see if the estimated decision was returned. Again, check for other places where this applies.

s-laugh (changes requested):

lib/api/mapBenefits.js
```diff
@@ -67,7 +68,9 @@ export function MapEICard(eiBenefitData) {
       break
 
     case StatusCodes.applicationReceived:
-      const estimatedDateOfDecision = new Date(eiBenefitData.nextRptDueDate)
+      const estimatedDateOfDecision = new Date(
+        formatDate(eiBenefitData.publishDate, 'en')
```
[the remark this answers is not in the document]
TishamIslam: To be clear, I was doing this to add the `T12:00:00.000Z` to the date and then parsing it, to avoid inconsistencies with the shown date. Though, it now occurs to me that there's a better way to go about this, so I'll do that.

s-laugh (approved):
Just leaving one additional suggestion, and a note that we probably need to re-visit how the newNormalizeDate function works.

lib/api/mapBenefits.test.js
```diff
@@ -40,13 +54,13 @@ describe('maps EI', () => {
     expect(mappedData.summaries).toBeFalsy()
   })
 
-  it('calculates estimatedDecisionDate', () => {
-    const dateTT = '2023-04-24'
-    const dateTR = new Date('2023-05-24').toDateString() //plus 1 month
+  it('Calculates estimatedDecisionDate', () => {
+    const dateTT = '2060-04-24'
+    const dateTR = newNormalizedDate('2060-05-22').toDateString() //plus 28 days
```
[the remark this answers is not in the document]
TishamIslam: Good point!

PR #413, opened 2022-06-27
Dcec 1938 add AEM implementation

DaceyTom2:
### Description
add AEM implementation

List of proposed changes:

- GraphQL client added
- GraphQL dependencies added
- AEM endpoint configured
- Example of .graphql query commented out on Dashboard page

App has boiler plate code and should run as previously


s-laugh (commented):
I tried running this locally with the latest env from key valut but am getting errors in the console.
Also I think this would be a good PR to review with the team so we all better understand the nuances of what's happening and how we will be using it.

s-laugh (approved):

PR #414, opened 2022-06-27
DCWC 1944 - Added logic for greeting

CurtisUnderwood:
### Description

List of proposed changes:

- Logic for greeting based on Date() function
- Added test for rendering the greeting


components/molecules/Greeting.js
```diff
@@ -5,10 +5,25 @@ import fr from '../../locales/fr'
...
+  function getGreeting() {
+    switch (true) {
+      case props.time < 12:
+        return t.morningGreeting
+      case props.time < 17:
+        return t.afternoonGreeting
+      case props.time < 24:
+        return t.eveningGreeting
+      default:
+        return t.greeting
```
s-laugh, line 17: The logic seems sound, except this is probably best suited for an if statement then a switch.
As well, with AEM, this component won't have access to the `t` translation. This would be a good time to remove it altogether from this file and have the dashboard page call a lib function to pass the right text directly to the component. Should be very similar to how the UniversalBenefitCard component works.

components/molecules/Greeting.js
```diff
@@ -5,10 +5,25 @@ import fr from '../../locales/fr'
 export default function Greeting(props) {
   const t = props.locale === 'en' ? en : fr
 
+  function getGreeting() {
+    switch (true) {
+      case props.time < 12:
```
TishamIslam, line 10: You could add another case for `props.time < 6` if a user happens to be burning the midnight oil  
```suggestion
      case props.time < 6:
        return t.eveningGreeting
      case props.time < 12:
```

s-laugh (changes requested):

pages/dashboard.js
```diff
@@ -79,7 +81,17 @@ export default function Dashboard(props) {
 
   return (
     <>
-      <Greeting locale={props.locale} name="Mary" />
+      <Greeting
+        locale={props.locale}
```
s-laugh, line 85: locale isn't needed anymore

lib/Utils.js
```diff
@@ -20,3 +20,15 @@ export function formatMoney(money, locale) {
   })
   return CAD.format(money).replace('CA', '').trim()
 }
+
+export function getGreeting(time) {
```
s-laugh, line 24: this should get a test for each return

s-laugh (commented):
Looking good, I think it just needs a few touch-ups.

s-laugh (approved):
Great!

PR #415, opened 2022-06-27
Auth for dynamic branches

DaceyTom2:
### Description
Toggling auth on for dynamic branches

List of proposed changes:

- Auth enabled to test ECAS implementation
- App should still work as usual


s-laugh (approved):
👍

PR #418, opened 2022-06-28
Update profile page with new look

TishamIslam:
### Description
Changed the profile page to fit the new designs.

List of proposed changes:
- Removed the security parts of the page, alongside the sets of tabs.
- Changed up the `ProfileInfo` and `ProfileInfoSection` components to better reflect the new design for the profile page.


components/molecules/ProfileInfo.js
```diff
@@ -1,17 +1,27 @@
-import ProfileInfoSection from '../atoms/ProfileInfoSection'
+import en from '../../locales/en'
+import fr from '../../locales/fr'
```
s-laugh, line 2: We shouldn't be importing translations into components anymore, the text should be fed though props to support AEM.

locales/en.js
```diff
@@ -450,13 +450,17 @@ export default {
...
+  profileSettings: 'Profile settings',
+  updateProfile: 'Update your profile information',
   profileAndPreferences: 'Profile and Preferences',
   securitySettings: 'Security Settings',
   backToDashboard: 'Back to my dashboard',
   edit: 'Edit',
   moreInfo: 'More Information',
 
+  //Security page
+  securitySettings: 'Security settings',
```
s-laugh, line 462: FYI, LGTM noted this is duplicated

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
-          ) : profileTabSelected === 1 ? (
-            <ProfileInfo fields={[fakeFieldsCPP]} locale={props.locale} />
-          ) : profileTabSelected === 2 ? (
-            <ProfileInfo fields={[fakeFieldsOAS]} locale={props.locale} />
-          ) : null}
-        </>
-      ) : tabSelected === 1 ? (
-        <ProfileInfo fields={[fakeSecurityFields]} locale={props.locale} />
-      ) : null}
+      <hr className="border-1 border-red-400" />
```
s-laugh, line 122: hr shouldn't be used like this; adding a border to the h1 should do the trick.

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
+                <ProfileInfo
+                  fields={benefitInfo.fields}
+                  locale={props.locale}
+                />
+              ) : null}
+              <ViewMoreLessButton
+                expanded={benefitInfo.show}
+                icon={benefitInfo.show}
+                onClick={() => benefitInfo.setShow(!benefitInfo.show)}
+                caption={!benefitInfo.show ? 'View settings' : 'View less'}
```
s-laugh, line 140: make sure to get content from the translation files

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
+            </div>
+
+            <hr className="border-1 border-gray-500" />
+          </>
+        )
+      })}
+
+      <div className="mt-10">
+        <h2 className="text-3xl font-bold">Looking for security settings?</h2>
+        <ul className="list-disc ml-8 text-lg">
```
s-laugh, line 151: I think we should just us a p tag here and a a list like style; as there is only ever one item in the list.

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
+          <Link href={'/security-settings'}>
```
s-laugh, line 152: Don't forget to add the `passHref`
```suggestion
          <Link href={'/security-settings'} passHref>
```

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
-        </>
-      ) : tabSelected === 1 ? (
-        <ProfileInfo fields={[fakeSecurityFields]} locale={props.locale} />
-      ) : null}
+      <hr className="border-1 border-red-400" />
+      <span className="text-lg">{t.updateProfile}</span>
+
+      {benefitInformations.map((benefitInfo) => {
+        return (
+          <>
```
s-laugh, line 127: I think it might be a good idea to include all of the mapped Info into the ProfileInfo component (including the state change of the view more.

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
+            <h2 className="text-3xl py-2 font-bold">{benefitInfo.benefit}</h2>
+            <div className="py-2 mb-4">
+              {benefitInfo.show ? (
+                <ProfileInfo
+                  fields={benefitInfo.fields}
+                  locale={props.locale}
+                />
+              ) : null}
+              <ViewMoreLessButton
```
s-laugh, line 136: Double check the styling on this, it may need an option to override some styles.

s-laugh (changes requested):
Lots of notes and suggestions, but it looking good!

pages/profile.js
TishamIslam: Maybe a larger component that maps the benefit information as it does on the page instead? As the component is right now, you could put it on the security page without any changes required.

components/molecules/ProfileInfo.js
TishamIslam: Done.

pages/profile.js
TishamIslam: Nice catch

pages/profile.js
TishamIslam: Ah yeah, wasn't sure if I should've added something like that. Done, removed the left margin of the buttons on the page

pages/profile.js
TishamIslam: Not sure if that's something that we want at all though.

s-laugh: The security page is quite static, so it doesn't quit relate. We can re-visit later if need if you feel this is the better implementation for now.

pages/profile.js
s-laugh: The padding and margins in theses sections still seem quite off from the designs.

pages/profile.js
TishamIslam: Hmm, yeah sure let's revisit for now.

pages/profile.js
TishamIslam: Seems like the `pl-0` class doesn't apply when it comes after `pl-2`, so I removed the padding out of the component.

pages/profile.js
```diff
@@ -3,187 +3,164 @@ import Link from 'next/link'
...
+      setShow: setViewOASFields,
+      fields: fakeFieldsOAS,
+      benefit: t.oas,
+    },
+  ]
   return (
     <>
-      <h1 className="py-4 text-4xl font-bold text-gray-darker">
-        {t.profileAndSecuritySettings}
+      <h1 className="py-4 text-4xl font-bold text-gray-darker border-b border-red-400">
```
s-laugh, line 119: I didn't find this before, but we should be using the "heading" component from the design system for the h1 
https://dts-stn.github.io/DECD-Design-System/main/?path=/story/components-heading--default

s-laugh (commented):
Can you please take a close look at the designs for the whole page and ensure the spacings between elements are accurate; there looks to be quite a few inconstancies.

pages/profile.js
```diff
@@ -3,187 +3,165 @@ import Link from 'next/link'
...
+          </>
+        )
+      })}
+
+      <div className="mt-10">
+        <h2 className="text-3xl font-bold">Looking for security settings?</h2>
+        <ul className="list-disc ml-8 text-lg">
+          <Link href={'/security-settings'} passHref>
+            <li>
+              <a className="underline text-blue-600 cursor-pointer hover:text-blue-800 visited:text-purple-600">
```
s-laugh, line 155: Just one last thing, we should use the colours from our pallet and not tailwinds.
```suggestion
              <a className="underline text-blue-default cursor-pointer hover:text-blue-hover visited:text-purple-medium">
```

TishamIslam: Done. Should I also change the `border-gray-500` for the line separators? The closest colour in our palette (gray-dark) is fairly different compared to the actual design.

s-laugh: yes... you could use the HorizontalRule component which we have in other places, it's using border-gray-light
There's also a design system component we aren't using that does the same thing.

PR #421, opened 2022-06-30
1986 security settings page

s-laugh:
### Description

List of proposed changes:

- Adding the security page
- Making use of the action button in the info section with some secondary styles


nibivi77 (changes requested):
Extra small comment. All good otherwise!

nibivi77 (approved):
Works for me!

PR #423, opened 2022-07-04
Dcwc 1946 ordered benefits

nibivi77:
### Description

Ordering benefits as mentioned in the ticket.


pages/dashboard.js
```diff
@@ -45,39 +45,136 @@ export default function Dashboard(props) {
...
-          if (res.status === 200) {
-            setData(await res.json())
+    if (
+      !cppLoaded &&
+      !cppdLoaded &&
+      !eiLoaded &&
+      !oasLoaded &&
+      !gisLoaded &&
+      !sebLoaded
+    ) {
```
s-laugh, line 64: I'm wondering if these would be better placed around each of their own 'fetchProgramData' ? so that each program operates a little more independently.

pages/dashboard.js
```diff
@@ -45,39 +45,136 @@ export default function Dashboard(props) {
...
+  }, [cppLoaded, cppdLoaded, eiLoaded, gisLoaded, oasLoaded, sebLoaded])
+
+  useEffect(() => {
+    if (
+      cppLoaded &&
+      cppdLoaded &&
+      eiLoaded &&
+      oasLoaded &&
+      gisLoaded &&
+      sebLoaded
```
s-laugh, line 100: I'm guessing this will block results showing on the page until all programs have returned data, is that right?
_It's probably okay to leave for now, but just want to be aware of it._

pages/dashboard.js
```diff
@@ -45,39 +45,136 @@ export default function Dashboard(props) {
...
+      let orderedBenefitUpdate = []
+      let orderedInPayment = []
+      let orderedExhausted = []
+      let orderedInactive = []
+      let orderedActiveAgreement = []
+
+      newArray.forEach((benefits) => {
+        if (benefits != undefined) {
+          benefits.forEach((benefit) => {
+            switch (benefit.statusCode) {
```
s-laugh, line 123: This is a really neat concept for sorting, I like, just wondering if you chose not to use `.sort(func)` for a particular reason?

pages/dashboard.js
```diff
@@ -92,187 +189,36 @@ export default function Dashboard(props) {
         myBenefitsAndServices={t.myBenefitsAndServices}
       />
       <div className="mb-8">
-        {cppLoaded ? null : 'Loading CPP User Benefit Data...'}
-        {cppError ?? null}
```
s-laugh, line 96: I think it will be helpful to keep the unstyled loading and error text for the time being.

s-laugh (approved):
Awesome work!
No blocking comments, just getting a sense for your thinking around the implementation.

pages/dashboard.js
nibivi77: I see what you mean, if you prefer that, we can have it that way. Say the word and I'll make the change.

pages/dashboard.js
nibivi77: Yes exactly. It gets really tricky handling it seperately and making all benefits show as soon as they are loaded individually. It would also make the page jump a lot if we display them as soon as some are loaded.

The current way its implemented will load them all together. Makes the page 'jump' only once.

s-laugh: Yeah, we'll need a business decision on this at some point. If we do wait for all the data before showing this could be something we move to our API at some point.

pages/dashboard.js
s-laugh: All good, using sort could be quite complex as it's all about evaluating the items within an array returning a positive or negative for each pair. As it's generally just used for numbers or alphabetical lists, using the custom override for something like this wouldn't be straight forward, what you've go here is easy to read.

pages/dashboard.js
s-laugh: You're the author on this, can leave it as it. It's consistent with the other useEffect.

PR #424, opened 2022-07-05
style update for ad cards

shewood:
### Description

List of proposed changes:
Updated the styling on the advertising cards to match Figma
-
-

### DIFF
--- a/components/molecules/BenefitApplicationCard.js
+++ b/components/molecules/BenefitApplicationCard.js
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types'
 import HorizontalRule from '../atoms/HorizontalRule'
 import propTypes from 'prop-types'
 import { ProgramCodes } from '../../constants/ProgramCodes'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faStamp } from '@fortawesome/free-solid-svg-icons'
 
 const BenefitApplicationCard = (props) => {
   const t = props.locale === 'en' ? en : fr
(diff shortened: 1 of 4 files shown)


components/molecules/BenefitApplicationCard.js
```diff
@@ -26,42 +28,36 @@ const BenefitApplicationCard = (props) => {
...
         >
-          <h2 className="font-bold font-display text-4xl sm:text-lg md:text-xl lg:text-3xl xl:text-4xl">
+          <h2 className="font-bold font-display text-4xl sm:text-6xl">
             {benefitName}
           </h2>
-          <div className="font-bold font-display text-md sm:text-md lg:text-md mb-2 w-44 sm:w-32 lg:w-44">
+          <div className="font-bold font-display text-md sm:text-md lg:text-md mb-2 w-44 sm:w-32 lg:w-44 ">
             {t[benefitSubType]}
           </div>
+          <ul className="pb-6 sm:mb-7">
```
nibivi77, line 43: I don't think we need a list here. Leaving just the `<p>` should be fine. 

This might get confusing to the screen readers.

nibivi77 (commented):
Looks good to me!

Just a small comment.

Also, we are not using the same icons as the design. I assume that's because we could not extract the image from Figma?

components/molecules/BenefitApplicationCard.js
shewood: I copied the the universal card for consistency, i am fine either way.

s-laugh: I'm not sure just the p tags is best, there should be some kind of association between the two texts. The status acting like a "label".

Just some context: It was originally suggested we use a data-definition-list for the summaries but because of the links in some, we opted for a regular unordered list; that patter is copied over here.

shewood: OK, so we should leave it a list then?

shewood:
@nibivi77 regarding the icons. Yes, these are the free icons to use. I checked with Claudia. She might change them later.

PR #432, opened 2022-07-07
Font Awesome Library across app

DaceyTom2:
### Description
Added Font Awesome Library across app

List of proposed changes:
- Can now get a string from AEM and use it for fonts. 
eg. <FontAwesomeIcon icon="fa-solid fa-file-lines" />

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -2,6 +2,19 @@ import '../styles/globals.css'
 import '../styles/fonts.css'
 import Layout from '../components/organisms/Layout'
 import { useRouter } from 'next/router'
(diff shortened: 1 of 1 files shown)


s-laugh:
I was thinking it would be better to go with dynamic icon importing instead of globally.

s-laugh:
> @s-laugh Do you have an example or link?

https://fontawesome.com/v6/docs/web/use-with/react/add-icons#dynamic-icon-importing

Should still let us specify an icon by a string name.
We'd specify the style (solid, regular, brands, ext..) in the code and have to add the babel-plugin-marcos lib.

PR #435, opened 2022-07-07
add dependabot config

s-laugh:
### Description

List of proposed changes:

- Adding a dependabot config to change the branch separator to `-`.
  - This is in hopes that it would get the branch deployments to work but the names might still be too long.

We'll probably just have to wait for the bot to flag and open a PR.


nibivi77 (approved):
Only way to see if it works is by approving it right?

PR #450, opened 2022-07-12
Readme update

s-laugh:
### Description

List of proposed changes:

- Remove forced colour on badges so errors show red
- Added coverage badges (will only work after this change is merged, as it's looking for the coverage-summary.json in dev that doesn't exist yet)
- Added some version badges

### Additional Notes

Copied from next-template


shewood (approved):
Nice!

PR #451, opened 2022-07-12
Added advertisement card logic

TishamIslam:
### Description
Adds logic to the advertisement cards, as shown below, from [here](https://014gc.sharepoint.com/:x:/r/sites/DECD/_layouts/15/doc2.aspx?sourcedoc=%7BABFCF0A7-6257-4D36-AF51-BC2C2912EBC0%7D&file=Backlog%20Overview.xlsx&action=default&mobileredirect=true&cid=1e73fc3a-aac0-4dcd-b4dc-31fc89ede769) such that they'll now conditionally render based on what benefit cards are already displayed. 
image

List of proposed changes:
- Conditionally display advertisement cards
-


contents/BenefitAdvertisingCards.js
```diff
@@ -143,21 +141,57 @@ const APPLICATION_CARD_CPP_DEATH_BENEFIT = {
...
-    APPLICATION_CARD_CPPD,
-    APPLICATION_CARD_CPP_CHILDS_BENEFIT_AGED_18_25,
-    APPLICATION_CARD_CPP_SURVIVOR_PENSION_AND_CHILD_BENEFITS,
-    APPLICATION_CARD_CPP_ALLOWANCE_OR_ALLOWANCE_FOR_SURVIVOR,
-    APPLICATION_CARD_CPP_PENSION_SHARING,
-    APPLICATION_CARD_CPP_CREADIT_SPLIT,
-    APPLICATION_CARD_CPP_CHILD_REARING_PROVISION,
-    APPLICATION_CARD_CPP_DEATH_BENEFIT,
-  ]
+function getAdvertisingCards() {
```
s-laugh, line 144: We'd want this function to simulate a result from AEM, and the prop `displayFlag` wouldn't be included in that.

pages/dashboard.js
```diff
@@ -117,9 +119,17 @@ export default function Dashboard(props) {
       let orderedInactive = []
       let orderedActiveAgreement = []
 
+      //deep clone advertisingCards, set displayFlag for true for now
+      const newAdCards = {}
```
s-laugh, line 123: It might be easier to just access the list from the props directly rather than re-creating the list.

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
...
+              newAdCards.EI.displayFlag = false
+            }
+            if (
+              benefit.programCode === ProgramCodes.CPP &&
+              benefit.typeCode === TypeCodes.CPPRetirement
+            ) {
+              //retirement benefit
+              newAdCards.CPP.displayFlag = false
+            }
+            if (benefit.typeCode === TypeCodes.OASRetirement) {
```
s-laugh, line 174: should check the program here too

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
...
+              newAdCards.OAS.displayFlag = false
+            }
+            if (benefit.typeCode === TypeCodes.GISRetirement) {
```
s-laugh, line 177: again, should check the program

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
...
+              newAdCards.GIS.displayFlag = false
+            }
+            if (benefit.programCode === ProgramCodes.CPPD) {
```
s-laugh, line 180: should check the type as well

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
...
+              newAdCards.CPPD.displayFlag = false
+            }
+            if (false) {
```
s-laugh, line 183: this needs a condition

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
...
+              //No CPP benefit card with Child's Benefit implemented
+              newAdCards.CPP_child_benefit_aged_18_25.displayFlag = false
+            }
+            if (
+              benefit.programCode === ProgramCodes.CPP &&
+              benefit.typeCode === TypeCodes.CPPSurvivor
+            ) {
+              newAdCards.CPP_survivors_pension_and_childrens_benefits.displayFlag = false
+            }
+            if (false) {
```
s-laugh, line 193: This also needs a condition

pages/dashboard.js
```diff
@@ -146,10 +156,50 @@ export default function Dashboard(props) {
                 orderedActiveAgreement.push(benefit)
                 break
             }
+
+            //determine what benefits to advertise
```
s-laugh, line 160: there is a fair bit of repetitive code here, see if there are ways to reduce the repetitiveness so it's easy to expand on.

s-laugh (changes requested):
Your logic looks to still need some work; and it would be wonderful if there were tests added for each condition you add.

pages/dashboard.js
TishamIslam: Tried a different approach, haven't added tests yet though. Not sure about my approach for verifying program codes since all it does is verify that a benefit has a correct code, but I don't think there's an alternative approach that verifies that the  program code matches with the typecode without un-generalizing it. 

Side note, I noticed that atm CPPD benefits all fall under the type of `CPPRetirement` which, assuming it isn't intentional, means that CPPD advertisement cards will continue to show even if there's CPPD benefit cards present.

s-laugh: Some of the separation of concerns is starting to show with this, which is great, but more can be done to make the code more readable.
Also, I think you've started to over complicated the logic, it should pretty easy to follow - the logic itself isn't that complicated. Each application card has it's own matching conditions based on Program, sometimes Type and sometimes Status. While each condition is unique, the way we've set up the benefit objects is so they are not unique, so use that!

s-laugh (commented):
This is coming along :)

contents/BenefitAdvertisingCards.js
```diff
@@ -1,7 +1,40 @@
...
+  CPPchild_benefit_aged_18_25: true, //Always, unless, any CPP benefit card is shown with the benefit type Child's Benefit
+  CPPsurvivors_pension_and_childrens_benefits: true, //Always, unless, any CPP benefit card is shown with the benefit type survivors
+  CPPallowance_or_allowance_for_survivor: true, //Always, unless, any GIS benefit card is shown with the benefit type Allowace or Allowance for the Survivor
+  CPPpension_sharing: true, //Always
+  CPPcredit_split: true, //Always
+  CPPchild_rearing_provision: true, //Always
+  CPPdeath_benefit: true, //Always
+}
+
+function mapTypeCodesToFlags(typeCode) {
```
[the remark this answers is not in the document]
TishamIslam: I moved this function and also the switch case to determine ad flags out into a separate file. Will write tests for the latter soon.

contents/BenefitAdvertisingCards.js
```diff
@@ -1,8 +1,21 @@
-import en from '../locales/en'
-const t = en
 const applyIcon = '/images/dashboard/apply-for-benefit-icon.svg'
 const estimateIcon = '/images/dashboard/estimate-retirement-income-icon.svg'
 
+const defaultDisplayFlags = {
```
s-laugh, line 4: Have you considered creating property names that are representative of the program and type?
Something that works like TaskGroups https://github.com/DTS-STN/sc-digital-centre/blob/d44715a77e2713017d1b9ec108f1e98eb1777b10/contents/BenefitTasksGroups.js#L513

lib/mapAdCards.test.js
```diff
@@ -0,0 +1,126 @@
...
+      StatusCodes.inactive,
+      TypeCodes.EISickness
+    )
+    determineAdCards(eiBenefit, adCards)
+    expect(adCards.EI).toEqual(false)
+  })
+
+  it('EI - All statuses except inactive keeps EI flag as true', () => {
+    Object.entries(StatusCodes).forEach(([statusCode, statusString]) => {
+      if (statusCode !== 'inactive' || statusString !== StatusCodes.inactive) {
```
s-laugh, line 35: there is no reset of adCards, so this function may not always work as expected. I would suggest separating this into multiple tests. A test should test one thing at a time; single input, single output.

s-laugh (changes requested):
There is good progress here, I still have some concerns however.

lib/mapAdCards.js
```diff
@@ -0,0 +1,45 @@
+import { TypeCodes } from '../constants/ProgramTypeCodes'
+import { StatusCodes } from '../constants/StatusCodes'
+import { ProgramCodes } from '../constants/ProgramCodes'
+
+export function determineAdCards(benefit, adFlags) {
```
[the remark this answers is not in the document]
TishamIslam: Alright. Should I also add `adFlags = determineAdCards(benefit, adFlags)` wherever I call it?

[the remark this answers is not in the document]
s-laugh: yes, would recommend that.

contents/BenefitAdvertisingCards.js
TishamIslam: Added a `typeCode` property to each adcard, and changed the defaultDisplayFlag object to use ProgramCodes and TypeCodes. Also added some extra TypeCodes for the various CPP subtypes that are advertised.

pages/dashboard.js
```diff
@@ -150,10 +156,14 @@ export default function Dashboard(props) {
                 orderedActiveAgreement.push(benefit)
                 break
             }
+            console.log(benefit)
```
s-laugh, line 159: This shouldn't get merged.

s-laugh (approved):
Excellent work! Just one tiny change before you merge.

pages/dashboard.js
TishamIslam: Oops haha

PR #457, opened 2022-07-13
DCWC-1613 Build Death Benefit Card

CurtisUnderwood:
### Description

List of proposed changes:

- Added new status Paid for death benefit card
- Added Death Benefit Card (Paid, Application Received)

### Additional Notes
- Existing issue with Task Groups that might require a second look (A specific TypeCode CPPDeath needs its own set of tasks outside of CPP?)


s-laugh (commented):
Looks great! Would love to see some tests.

PR #466, opened 2022-07-14
Dcwc 2010 decision date

TishamIslam:
### Description
Application received will now always show the estimated decision date.

List of proposed changes:
- Removed logic to hide estimated decision date 
-


lib/api/mapBenefits.js
```diff
@@ -81,16 +81,11 @@ export function MapEICard(eiBenefitData) {
           eiBenefitData.publishDate,
           eiBenefitData.messageId
         ),
+        FormatSummary(
+          SummaryTypes.EstimatedDecisionDate,
+          estimatedDateOfDecision.getTime()
+        ),
       ]
-      //remove estimated decision date if current day is past that point
-      if (Date.now() < estimatedDateOfDecision.getTime()) {
```
s-laugh, line 86: I don't believe the fix was to remove this logic but to update the dates so they show... should confirm with Breanna

s-laugh: Discussed at stand-up; check notes in ticket. Ask Breanna if clarification still needed.

s-laugh: I don't think bumping every date is a good idea, a lot of it is in place to demo a scenario, you should confirm with Breanna if she's okay with that. My understanding is just the dates that effect the display of "Estimated Decision" for the "Application Revived" should change.

s-laugh:
The latest build seems to be hung, and unable to load.

lib/api/mapBenefits.js
TishamIslam: Breanna said she's okay with bumping every date up

s-laugh (approved):
Good work!

PR #472, opened 2022-07-15
adding names to workflows

s-laugh:
### Description

List of proposed changes:

- Update naming for tests workflows


CurtisUnderwood (changes requested):
Seems like there's an error causing dast to fail?

image

s-laugh:
> Seems like there's an error causing dast to fail?
Good catch!

CurtisUnderwood (approved):
Nice!

PR #473, opened 2022-07-15
Hook-up Introp prod-like API payload for CPP & EI

s-laugh:
### Description

List of proposed changes:

- Configure the CPP and EI mapings for Introp expected prod-like payload (prop name changes)
- Updates mock data to match new formating
- Handles new messaging format from EI


nibivi77 (approved):
I've been reviewing this PR for a while now and I can't seem to see something to change. 

Check the conflicts this PR has and it should be good!

PR #483, opened 2022-07-19
Lint clean

s-laugh:
### Description

List of proposed changes:

- Cleans up lint errors
  - Uses next image in splash page
  - Moves `beforeInteractive` script to _document as required by next (adobe implementation will likely need debugging when properly implemented)
  - Set's some boolean properly into props
- Removes a number of un-used components (used design system components)
  - Keep the DSHeader as it contains logic, remove DSFooter as it's just a pass through

Everything still works exactly like it did before

### Additional Notes

Removal of the components is being be discussed as a team in slack.


nibivi77 (approved):
I really like this PR Shaun. Clean and well maintained.

PR #498, opened 2022-07-26
63866 Confirm connection with ECAS

DaceyTom2:
### Description
Authentication with ECAS

List of proposed changes:
- Configured Next-auth library

Login to ECAS and view Dashboard page

### Additional Notes
Next-auth forked to allow nonce in auth/token request

To be added in a later PR
- ECAS to return to dashboard page
- SSL enabled
- Bypassing the providers page


lib/api/programData.js
```diff
@@ -23,10 +24,14 @@ export async function FetchProgramData(
 ) {
   switch (req.method) {
     case 'GET':
-      if (!AuthIsDisabled() && !(await AuthIsValid(req)))
+      if (
+        !AuthIsDisabled() &&
+        !(await AuthIsValid(req)) &&
+        !(await getToken({ req }))
```
s-laugh, line 30: AuthIsValid is using the getSession should that be change to getToken?
Just seems like there's some duplicate validation here. Also maybe only need to getToken once?

pages/api/auth/[...nextauth].js
```diff
@@ -1,41 +1,62 @@
...
-    */
     {
-      id: 'customProvider',
-      name: 'customProvider',
+      id: 'ecasProvider',
+      name: 'ECAS',
       clientId: process.env.CLIENT_ID,
-      clientSecret: process.env.CLIENT_SECRET,
       type: 'oauth',
-      wellKnown: process.env.WELL_KNOWN,
```
Comment (s-laugh) [inline pages/api/auth/[...nextauth].js:19]: If we are removing key cloak, what do you think about removing the env variables for it as well (docker/team city files)? Maybe at least give them less generic names.

s-laugh (approved):

PR #500, opened 2022-07-27
Expand/Collapse Card Interaction

CurtisUnderwood:
### Additional Notes
- Added transition for expansion/collapse of universal benefit card
-
## Checklist

- [x] Deployment working

### DIFF
--- a/components/molecules/UniversalBenefitCard.js
+++ b/components/molecules/UniversalBenefitCard.js
@@ -92,7 +92,7 @@ export default function UniversalBenefitCard(props) {
       </h3>
       <div className=" flex flex-col bg-white  rounded-b-xl">
         {props.taskGroups == null || props.taskGroups.length <= 0 ? null : (
-          <div className="  ">
+          <div id={taskListId}>
             {!isOpen ? null : (
               <div className="pb-12">
                 <div className="bg-gray-lighter grid grid-rows-1 md:grid-cols-2">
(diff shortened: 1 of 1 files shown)


TishamIslam:
Not sure what happened, but I think it's having a weird interaction with the EI card since it has two sections of tasks.
image

After expansion it behaves normally.

shewood (changes requested):
I see the same issue as Tisham but only on Firefox and only for that card.

shewood:
I think the only change you need to make on the previous version is this to get it to do what we want.  
Change 
 block: 'start',
to
 block: 'nearest',

CurtisUnderwood:
> I see the same issue as Tisham but only on Firefox and only for that card.

Good catch, seemed to be an issue only on Firefox can you confirm it's resolved?

> I think the only change you need to make on the previous version is this to get it to do what we want.

I think this may be a question for the PO, the card expansion ticket asks for a transition animation on the card dropdown which required a bit of a re-organization. Block 'nearest' would just remove the scroll, both are viable just depends if we want the animation.

components/molecules/UniversalBenefitCard.js
```diff
@@ -97,7 +92,7 @@ export default function UniversalBenefitCard(props) {
       </h3>
       <div className=" flex flex-col bg-white  rounded-b-xl">
         {props.taskGroups == null || props.taskGroups.length <= 0 ? null : (
-          <div id={taskListId} className="  ">
+          <div className="  ">
```
shewood, line 95: Looks good.
Any reason for not keeping the id on the tasklist?

PR #509, opened 2022-08-02
63866 AEM integration

nibivi77:
### [ADO Work Item 63866](https://dev.azure.com/VP-BD/DECD/_workitems/edit/63866)

This PR is a first step to get connected to AEM. For now, the query only fetches
 ` scPageNameEn
      scPageNameFr
      scTitleEn
      scTitleFr`
.
## Checklist

- [x] Tests added for key features and bugs
- [ ] Deployment working TO BE TESTED

### DIFF
--- a/graphql/queries/dashboardQuery.graphql
+++ b/graphql/queries/dashboardQuery.graphql
@@ -1,12 +1,12 @@
 query getDashboardPage {
-	scLabsErrorPageByPath(_path: "/content/dam/decd-endc/content-fragments/alpha/sclabs/pages/500-error-page") {
-	    item {
-            sclContentEn {
-                json
-            }
-            sclContentFr {
-                json
-            }
-        }
-	}
+  alphaSCHPageByPath(
+    _path: "/content/dam/decd-endc/content-fragments/alpha/secure-client-hub/pages/dashboard"
+  ) {
+    item {
+      scPageNameEn
+      scPageNameFr
+      scTitleEn
+      scTitleFr
+    }
+  }
 }
--- a/lib/aem/mapper.js
+++ b/lib/aem/mapper.js
@@ -1,0 +1,19 @@
+import queryGraphQL from '../../graphql/client'
+import getDashboardPage from '../../graphql/queries/dashboardQuery.graphql'
+
+export default async function MapDashboard() {
+  const aemContent = await queryGraphQL(getDashboardPage).then((result) => {
+    return result
+  })
+  const returnedQuery = aemContent.data.alphaSCHPageByPath.item
+  return {
+    en: {
+      pageName: returnedQuery.scPageNameEn,
+      pageTitle: returnedQuery.scTitleEn,
+    },
+    fr: {
+      pageName: returnedQuery.scPageNameFr,
+      pageTitle: returnedQuery.scTitleFr,
+    },
+  }
+}
--- a/lib/aem/mapper.js
+++ b/lib/aem/mapper.js
@@ -2,18 +2,25 @@ import queryGraphQL from '../../graphql/client'
 import getDashboardPage from '../../graphql/queries/dashboardQuery.graphql'
 
 export default async function MapDashboard() {
-  const aemContent = await queryGraphQL(getDashboardPage).then((result) => {
-    return result
-  })
-  const returnedQuery = aemContent.data.alphaSCHPageByPath.item
-  return {
-    en: {
-      pageName: returnedQuery.scPageNameEn,
-      pageTitle: returnedQuery.scTitleEn,
-    },
-    fr: {
-      pageName: returnedQuery.scPageNameFr,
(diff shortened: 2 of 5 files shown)


graphql/queries/dashboardQuery.graphql
```diff
@@ -1,12 +1,12 @@
...
-}
\ No newline at end of file
+  alphaSCHPageByPath(
+    _path: "/content/dam/decd-endc/content-fragments/alpha/secure-client-hub/pages/dashboard"
+  ) {
+    item {
+      scPageNameEn
+      scPageNameFr
+      scTitleEn
+      scTitleFr
```
s-laugh, line 9: hmm, I don't think this is an ideal formatting...
This requires us to build/maintain that mapping function (`MapDashboard`) just so we can toggle the language - that's going to be very error prone.
A preferable method, in my opinion, is one where we don't need to "map" the objects returned from AEM. While we don't control objects from Interop, we should from AEM. We can edit the application to interpret the new AEM object (rather than the current one), but it needs to enable us to toggle language.

Again, any change here would require AEM change too, but I see this a bit more as a "blocker".

lib/aem/mapper.js
```diff
@@ -0,0 +1,26 @@
...
+        pageName: returnedQuery.scPageNameEn,
+        pageTitle: returnedQuery.scTitleEn,
+      },
+      fr: {
+        pageName: returnedQuery.scPageNameFr,
+        pageTitle: returnedQuery.scTitleFr,
+      },
+    }
+  } catch (e) {
+    return {
```
s-laugh, line 21: **Blocking** - Returning an object like this (one with a format that doesn't match the expected result) would cause the application to crash at runtime. I suspect if AEM fails we won't have a backup for content, so we'd want to redirect to a 500 error page and log the error.

s-laugh (changes requested):
Only my last comment is **blocking**. The other two are far more opinion based, so not-blcoking. If you don't feel strongly either way, probably good to chat with the team in slack about it. :slightly_smiling_face:

graphql/queries/dashboardQuery.graphql
```diff
@@ -1,12 +1,12 @@
...
-                json
-            }
-            sclContentFr {
-                json
-            }
-        }
-	}
-}
\ No newline at end of file
+  alphaSCHPageByPath(
```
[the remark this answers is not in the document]
DaceyTom2: @s-laugh discuss with Izabelle on naming conventions.
This isn't in reference to SCH being in alpha but rather the first version of the content model

lib/aem/mapper.js
DaceyTom2: The Dashboard page should be responsible for redirecting when an error is returned that can't be handled

graphql/queries/dashboardQuery.graphql
DaceyTom2: This is absolutely critical and @krischarbonneau  with SCLabs is planning to follow suit. We have found in the past that this is less error prone than trusting the CMS response

graphql/queries/dashboardQuery.graphql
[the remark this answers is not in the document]
s-laugh: continuing conversation in slack https://platypus-sc.slack.com/archives/C02GRL93BGT/p1659536530459419

graphql/queries/dashboardQuery.graphql
s-laugh: As agreed in the meeting, we will keep the mapping.

pages/dashboard.js
```diff
@@ -290,10 +289,17 @@ export async function getServerSideProps({ req, res, locale, query }) {
...
   setCookie('userid', userid, { req, res, maxAge: 60 * 6 * 24 })
 
-  // const aemContent = await queryGraphQL(getDashboardPage).then((result) => {
-  //   return result;
-  // });
-  // console.log(aemContent)
+  // Get mapped content from AEM
+  let aemContent
+  try {
+    aemContent = (await getDashboardContent()) || {}
```
s-laugh, line 295: If you want this to be a bit cleaner (aka, not changing the code for a test); in the test you should be able tell the mock what it should return for this function. Something like this
```js
jest.mock('../../lib/aem/mapper', () => {
   getDashboardContent: () => {}
 })
 ```

s-laugh (approved):
Great work! Just leaving an alternative suggestion for the mock.

pages/dashboard.js
nibivi77: I can't seem to get to work in this way. I keep getting errors. Did you want to give it a try?

s-laugh: Gave it a shot. I led you the wrong way it seems... sorry.
All you should need to do is add this to the top of the failing test, then you can remove the OR condition.
`getDashboardContent.mockReturnValueOnce({})`

PR #513, opened 2022-08-03
Bug fix: client match server

DaceyTom2:
## Description of Changes
An warning occurred due to filtering no benefit cards without using useEffect

Open the console for /dashboard?userid=default
The console should no longer warn about div mismatches between server and client


shewood (approved):
Apart from the unused variable warning by LGTM, this looks good.
I no longer see the error in the console.

DaceyTom2:
Fixed unused import

PR #517, opened 2022-08-04
Added keyboard navigation to tasklist dropdown

TishamIslam:
Verify that keyboard functionality works as described.

### Additional Notes
Keyboard navigation should work as follows: 
Dropdown button can be used as normal (tab to it and hit enter/space)
Can tab to a tasklist and navigate: 
- between separate tasklists with left/right arrow keys or tab/shift+tab
- to first task in the list by pressing enter, or up/down arrow key
  - can go to next task in the list with right/down arrow keys
  - can go to previous task in the list with left/up arrow keys
  - can return to the tasklist by 
    - pressing escape
    - attempting to go to the previous task when you're at the first task
    - attempting to go to the next task when you're at the last task 

## Checklist

- [ ] Deployment working

### DIFF
--- a/components/molecules/BenefitTasks.js
+++ b/components/molecules/BenefitTasks.js
@@ -3,8 +3,43 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import Link from 'next/link'
 
 export default function BenefitTasks(props) {
+  function returnToTaskList() {
+    document.getElementById(props.indexedTaskListId).focus()
+  }
+
+  function handleTaskListItemNavigation(e) {
+    e.stopPropagation()
+    const parent = e.target.parentElement
+
+    switch (e.code) {
+      case 'ArrowUp':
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Great work Tisham, I can't find something to change :+1:

s-laugh (commented):
The bug is about drop down menu's; seemingly specifically the "menu" in the header (which design system would need to fix). @shewood is that right?

In testing I personally find the keyboard navigation of these lists very confusing for a few reasons
- Tabbing to a list group that is not actionable (clickable) is very unusual and tabs over the links; only way to get to a link is via arrows
- Once on a link, up arrow does a reverse behaviour (bounces back a forth between first link and list group)
I'm not sure if this is wrong or not, just pointing it out to be validated.

TishamIslam:
Misunderstood the issue, closing PR.

PR #519, opened 2022-08-05
67653 Test flow prep work mocking

DaceyTom2:
## Description of Changes
An example of mocking in tests that we should be working towards for cleaner unit testing

Ensure code and tests still work

### Additional Notes
Had to comment out a significant number of mapBenefits tests. They are better suited for cypress e2e testing


shewood (approved):
Looks good.

PR #543, opened 2022-08-16
ZAP Error Fixes - Define CSP

CurtisUnderwood:
- This PR defines our Content Security Policy and uses a nOnce to enforce a strict script policy
- Should resolve some of our ZAP issues around missing CSP

### Additional Notes
- This may need to be revisited

## Checklist

- [ ] Deployment working


__tests__/pages/_document.test.js
```diff
@@ -1,9 +1,16 @@
 import '@testing-library/jest-dom'
 import MyDocument from '../../pages/_document'
+import * as utils from '../../lib/Utils.js'
```
s-laugh: Suggestion: We typically import only the functions we need, not the whole library
```suggestion
import { generateNonce } from '../../lib/Utils'
```

__tests__/pages/_document.test.js
```diff
@@ -1,9 +1,16 @@
...
 
 describe('_document', () => {
+  jest.spyOn(utils, 'generateNonce').mockReturnValue('test')
```
s-laugh: Suggestion: This seems to be the more consistent way we've mocked tests in our apps
```suggestion
  jest.mock('../../lib/Utils', () => ({
    generateNonce: jest.fn().mockReturnValue('test'),
  }))
```

lib/Utils.js
```diff
@@ -39,3 +39,9 @@ export function getGreeting(time) {
     return 'greeting'
   }
 }
+
+import crypto from 'crypto'
```
s-laugh: Suggestion: Should move this import to the top of the file.

pages/_document.js
```diff
@@ -1,27 +1,54 @@
 import Document, { Html, Head, Main, NextScript } from 'next/document'
+import { generateNonce } from '../lib/Utils.js'
```
s-laugh: Suggestion: you don't actually need the `.js` when importing
```suggestion
import { generateNonce } from '../lib/Utils'
```

__tests__/pages/_document.test.js
nibivi77: jest.spyOn() is perfect on if we want to keep track of when generateNonce is being called. [Documentation](https://jestjs.io/docs/jest-object#mock-functions)

s-laugh: No problem, if you think it's better to keep as is; however that benefit isn't getting used in the test. It also seems some of that call tracking is enabled on .mock now... can see what Tom's example with it was https://github.com/DTS-STN/sc-digital-centre/blob/dev/lib/api/mapBenefits.test.js#L35

pages/_document.js
```diff
@@ -1,27 +1,54 @@
...
+              __html: `window.__webpack_nonce__ = "${nonce}"`,
+            }}
           ></script>
           {process.env.NEXT_PUBLIC_ADOBE_ANALYTICS_URL ? (
             <Script
+              id="1"
               strategy="beforeInteractive"
               src={process.env.NEXT_PUBLIC_ADOBE_ANALYTICS_URL}
+              nonce={nonce}
+              dangerouslySetInnerHTML={{
```
nibivi77: It would be good if we can prevent using dangerouslySetInnerHTML.

One solution would be to purify the element with [DOMPurify](https://github.com/cure53/DOMPurify)

Found it on StackOverflow [here](https://stackoverflow.com/questions/29044518/safe-alternative-to-dangerouslysetinnerhtml)

Option 2: [Interweave](https://www.npmjs.com/package/interweave)

If the team does not mind using `dangerouslySetInnerHTML`, I am ok with it.

pages/_document.js
```diff
@@ -1,27 +1,54 @@
...
+    csp += `style-src 'self' https://fonts.googleapis.com 'unsafe-inline';`
+    csp += `img-src 'self' data: blob:;`
+    csp += `font-src 'self' https://fonts.gstatic.com data:;`
+    csp += `frame-ancestors 'self';`
+
+    if (ctx.res) {
+      ctx.res.setHeader('Content-Security-Policy', csp)
+    }
+
+    return { ...initialProps, nonce }
```
DaceyTom2: If possible it is better to set headers in one spot. CSP might be getting overwritten in the next.config.js file

  {
    key: 'Content-Security-Policy',
    value:
      "default-src 'self'; base-uri 'self'; frame-ancestors 'self'; form-action 'self'; object-src 'none'; script-src-elem 'self'; script-src 'self' 'unsafe-eval'; connect-src 'self'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' https://fonts.gstatic.com; img-src 'self' data: https:",
  },

pages/_document.js
DaceyTom2: There are currently conversations about removing Adobe Analytics in favour of open telemetry. Might be worth removing this script all together

s-laugh (commented):
No additional comments from me. But agree with the others raised.

pages/_document.js
CurtisUnderwood: Seems like the HTML didn't have much of a purpose so it's removed.

pages/_document.js
CurtisUnderwood: Because some of the headers have a dynamic nonce or hash I believe that part needs to remain still in _document. I've moved our static CSP into the config as well and just appended script-src in _document but there are other ways to do this as well.

PR #547, opened 2022-08-16
Fix error thrown by access_token being null

TishamIslam:
## Description of Changes
Added optional chaining when attempting to get the access token, which should prevent run-time errors.   

## Checklist

- [ ] Deployment working


DaceyTom2 (approved):
Great catch

PR #548, opened 2022-08-16
Proxy configuration

DaceyTom2:
## Description of Changes
Automating part of the proxy configuration

Follow proxy documentation and ensure ECAS is accessible

### Additional Notes
Proxy configuration documentation added here
https://github.com/DTS-STN/sc-digital-centre/wiki/Proxy-Configuration

### DIFF
--- a/proxyconfig/client.js
+++ b/proxyconfig/client.js
@@ -1,0 +1,75 @@
+'use strict'
+
+Object.defineProperty(exports, '__esModule', {
+  value: true,
+})
+exports.openidClient = openidClient
+
+var _openidClient = require('openid-client')
+var HttpsProxyAgent = require('https-proxy-agent')
+
+async function openidClient(options) {
+  const provider = options.provider
+  let httpOptions = {}
+  if (provider.httpOptions) httpOptions = { ...provider.httpOptions }
+  if (process.env.http_proxy) {
+    let agent = new HttpsProxyAgent(process.env.http_proxy)
+    httpOptions.agent = agent
+  }
+  _openidClient.custom.setHttpOptionsDefaults(httpOptions)
+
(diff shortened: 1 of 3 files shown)


proxyconfig/client.js
```diff
@@ -0,0 +1,75 @@
+'use strict'
+
+Object.defineProperty(exports, '__esModule', {
+  value: true,
+})
+exports.openidClient = openidClient
+
+var _openidClient = require('openid-client')
```
nibivi77, line 8: `var` :grimacing:

proxyconfig/client.js
```diff
@@ -0,0 +1,75 @@
...
+          _provider$userinfo === void 0
+            ? void 0
+            : _provider$userinfo.url) !== null &&
+        _provider$userinfo$ur !== void 0
+          ? _provider$userinfo$ur
+          : provider.userinfo,
+    })
+  }
+
+  const client = new issuer.Client(
```
nibivi77, line 64: `const` should not be reassigned. (Line 73)

client[_openidClient.custom.clock_tolerance] = 10

nibivi77 (commented):
I know this code was taken from another source and it should not be modified but I left some suggestions.

proxyconfig/client.js
DaceyTom2: I'm going to leave these as var since it is an output from a typescript file module

proxyconfig/client.js
DaceyTom2: I'm going to leave these as var since it is an output from a typescript file module

nibivi77 (approved):
This works for me! The code is a little outdated but works well!

